PR #13, opened 2022-01-26
changed globals, that 15 weather regions can be choosed.

Hillenfz:
If user wants to simulate different weather than "Aachen", he has to change "location" in weather.py and pvs.py to of this locations:
"01Bremerhaven",
"02Rostock",
"03Hamburg",
"04Potsdam",
"05Essen",
"06Bad Marienburg",
"07Kassel",
"08Braunlage",
"09Chemnitz",
"10Hof",
"11Fichtelberg",
"12Mannheim",
"13Muehldorf",
"14Stoetten",
"15Garmisch Partenkirchen"


noah80 (changes requested):
No absolute paths to the desktop in public code please.

PR #21, opened 2022-02-17
Add default connections

joga4er:
(1) added default connections for building, heat pump, oil heater, districti heating and simple bucket boiler
(2) created example modular_household which represents all possible configurations of pv, boiler and heating systems
(3) manipulated hisim_main() to enable calls to modular_household_explicit( *arg ) with variable input arguments -> please check if this is ok for you Noah.
(4) extanded test_examples() by a loop over all possible configurations of modular_houshold


joga4er:
all relevant changes in add_default_connections_excluding_variable_inputs

PR #22, opened 2022-02-17
Add default connection excluding variable inputs

joga4er:
Same as "added-default-connections" but modular_houshold has no additional inputs and hisim.main remains unchanged. 
Automated tests for modluar household cannot be included at this stage

### DIFF
--- a/examples/basic_household.py
+++ b/examples/basic_household.py
@@ -7,6 +7,9 @@ from hisim.components import building
 from hisim.components import heat_pump
 from hisim.components import sumbuilder
 from hisim import utils
+
+import os
+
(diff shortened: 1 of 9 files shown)


joga4er:
(1) added default connections for building, heat pump, oil heater, districtict heating and simple bucket boiler
(2) created example modular_household which represents all possible configurations of pv, boiler and heating systems
(3) included dataclass SystemConfig to SimulationParameters class
(4) extanded test_examples() by a loop over all possible system configurations of modular_houshold -> tests take quite long now (roughly 15 minutes), maybe replace all configurations with random configuration in the future

noah80 (approved):

PR #58, opened 2022-06-23
Add some postprocessing

joga4er:
total consumption, production and injection as well as self consumption rate and autarky rate are written to report


noah80:
closed due to conflicts

PR #75, opened 2022-08-25
Include buffer storage

joga4er:
(1) generalized boiler to hot water storage which can be either used as buffer or boiler
(2) implemented heating system with buffer storage: was done with availbale controllers and components, component HeatingSystem is still not implemented
(3) separated example modular_household into functions initializing the corresponding components/HeatingSystems
(4) enabled parsing of SystemConfig and included decision variables included yes/no as well as capacity variables (5kWh)


noah80:
Redundant

PR #101, opened 2022-11-24
Building sizer progress

joga4er:
Building sizer operates with boolean and discrete options, but fitness function (kpi's) is not calculated and transfered well.


DavidNeuroth:
To avoid the mypy-error in building_sizer/setup.py, types-setuptools needs to be installed.
Should a requirements_dev.txt be created for such optional modules?

PR #120, opened 2023-02-10
Debugging again

joga4er:
(i) integrated increase of building temperature for clever control also in buffer case
(ii) solved bugs from UTSP tests
(iii) moved all cooling_considered variables when not needed anymore
(iv) made definition of car component optional


joga4er:
improoved control, so that buffer is automatically charged in surplus case, also when it is above lower hysteresis temperature

noah80 (approved):

PR #139, opened 2023-03-22
Advance kpis

joga4er:
(i) added buffer storage cycles and losses to report
(ii) fixed l1_heating_controller
(iii) fixed thermal mass values in report (calculation was right)
(iv) moved utsp url and api from system_config to archetype config


joga4er:
already included in other pull request

Issue #244, opened 2023-10-25
Add postprocessing option "System setup config to JSON"


k-knosala:
To test if the scaling works I would like to get the system setup configuration in postprocessing.

PR #248, opened 2023-10-31
Add `get_capacities` for postprocessing

k-knosala:
Since we are using the scaled defaults now, I need a way to get the scaled component sizes. This is why I added a `get_capacities()` method to the config base class.


k-knosala:
Solves #244

noah80:
will be implemented differently

PR #257, opened 2023-11-07
Renaming examples to system setups

k-rieck:


hisim/utils.py
```diff
@@ -49,11 +49,11 @@ def get_input_directory() -> str:
     ),
     "cfg": os.path.join(hisim_abs_path, "inputs", "cfg.json"),
     "utsp_results": os.path.join(hisim_results, "Results"),
-    "utsp_example_results": os.path.join(
+    "utsp_system setup_results": os.path.join(
```
k-knosala, line 52: This needs an underscore.

hisim/utils.py
```diff
@@ -49,11 +49,11 @@ def get_input_directory() -> str:
...
         hisim_inputs, "LPGResults_for_tests", "Results"
     ),
     "utsp_reports": os.path.join(hisim_results, "Reports"),
-    "utsp_example_reports": os.path.join(
+    "utsp_system setup_reports": os.path.join(
```
k-knosala, line 56: See above

k-knosala:
This also adds `setup_function` to all system setups and removes the argument `function_in_module` in hisim main.

k-knosala (commented):
Looks good to me apart from the two comments above.

hisim/utils.py
k-rieck: Thanks, I pushed again with an underscore

noah80 (approved):

PR #262, opened 2023-11-24
Changes on generic pv system

k-rieck:
- some changes on generic_pv_system.py
- so far two calculation methods:

1. simphotovoltaicfast (the one that was used so far, seems to have errors and ignores module and inverter data) 
2. simphotovoltaic_two (uses module and inverter data from database but apparently pvlib can only handle sandia modules)

the results of both methods seem similar although method 2. produces slightly less electricity:
-> method 2.- kpi production full year = 17602 kWh vs method 1.- kpi production full year = 18092 kWh
-> maybe more checks are needed
-> the link to the full year results will follow

- right now this branch uses still method 1.
- with method 1. (as no module and inverter data is read from the database) also more efficient cec modules as shown in the test  "test_generic_pv_system_fast_with_two_different_models.py" can be used


k-rieck:
link for full year simulaion results : https://drive.google.com/drive/folders/17EZjwnt50FTwznuvwrG0LCUdkSUVtzki?usp=drive_link

noah80 (approved):

PR #282, opened 2024-01-15
Unified system setup gas

k-knosala:
- Creates one system setup file for gas based households
- Uses Options class to toggle elements like PV, Cars, etc.


hisim/component.py
k-knosala, line 1: The only change in this file is in (old) lines 342-344. All other changes in this file are formatting.

noah80 (approved):

PR #293, opened 2024-01-24
Black formatting

k-knosala:
The line-length has not been consistennt througout the files. I am suggesting to use 120 as a default because 150 is to wide for side by side comparisons.

Formatted the folders `hisim` and `system_setups` with `black --line-length 120`.

Pycodestyle error E203 is disable according to the [Black documentation](https://black.readthedocs.io/en/stable/faq.html#why-are-flake8-s-e203-and-w503-violated).

Renamed `.pyproject.toml` to `pyproject.toml` to load the settings for black within the project.

We should transition the information in `setup.py` and `setup.cfg` to the `proproject.toml` because this is the [new standard for Python packaging](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html).

### DIFF
--- a/.prospector.yml
+++ b/.prospector.yml
@@ -46,6 +46,7 @@ pycodestyle:
     - N802
     - E128
     - E122
+    - E203
 
 pydocstyle:
   run: true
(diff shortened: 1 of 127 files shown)


k-knosala:
Merge #277 and #288 first to avoid conflicts.

PR #309, opened 2024-02-09
Fix energy performance for mass simus

k-rieck:
- problem was: 12% of all buildings in mass simulations (all building types and sizes) were too cold (Tmin < 14°C)
- the buildings that were too cold were the ones with high heating demands (elder, no refurbished houses)

- problem solution: scale heat pump power up for building heating demand over 130 kWh/m2a and even more for heating demands higher than 200 kWh/m2a (see advanced heat pump hplib)

- result: after the changes, no house was too cold anymore!

- sample tested for this purpose: 1000 randomly generated houses of all types and sizes


k-rieck:
heat pump scaling will not be implemented for now

PR #425, opened 2026-03-12
Multiple HiSim instances no longer interfere with each others' logging

Niteftef:
This branch / PR overhauls the logging. Instead of logging into the default path and moving the logs at the end of the simulation, it now temporarily stores any logs that are generated before the result directory is created, and as soon as that has happened, the stored logs and any further logs are immediately written into the result directory.
This way, the final log files are proper even if multiple instances are running at the same time.

The changes are mostly in log.py but also in hisim_main.py and simulator.py at the points where those functions set up logging.

The changes are minimally invasive, meaning that all other modules will continue working unchanged.


noah80:
Hi,

thank you for tackling the topic of the logging!

Kathi and me discussed the solution, but are not entirely convinced that that the minimal invasive approach is better than doing things correctly in the first place. Also, the idea of using globals for setting the log path would result in all houses in a co-simulated neighborhood ending the in the same log file. I'd much prefer avoiding globals if possible.
 
From my understanding we primarily need the pre-logging to debug a failing system setup.

So maybe a better approach would be to change the code so that the log file initialisation happens earlier instead of a rather complicated mechanism with multiple code paths? 

Best,
Noah

Niteftef:
Very valid. I tried meddle around as little as possible, hence this approach. But I agree that there are better ways (and using globals should be avoided whenever possible in any case).
So I take it I have the "green light" to go a bit deeper and will provide a solution including your feedback soon.

Niteftef:
Made a logger class and a kernel-specific object.
Due to being specific to the kernel instead of a simulation, it can be instantiated and is immediately available when log.py is first imported. For the same reason, the API does not change. log.warning() etc. still work fine.
To make it simulation-specific, Logger.setup() has to be called with the correct path.
To remove the bind to one specific simulation (f.ex. to run multiple simulations in the same kernel, sequentially), use Logger.reset().

I have moved my_sim.prepare_simulation_directory() to the simulator __init__(). Before, it was being called right after the init in hisim_main.initialize_from_json() but not at all in hisim_main.initialize_from_python(). To make up for the missing directory, it was called again in simulator.run_all_timesteps().
Calling it in simulator.__init__() seems to make more sense to me, but if there is important reasons why it needs to be elsewhere, just change it back.

Finally, I have removed code made redundant by setup() and reset().

Also, please fix prospector, I promise you it's not doing you any good in terms of codequality. Frankly, the state that it is in currently, it'd be better to just turn it off completely.

Niteftef:
I just remembered that default values can have some weird behavior, let me check if that's a problem and possibly write tests before merging.

Niteftef:
Fixed default value issue

k-rieck (approved):

PR #431, opened 2026-04-30
Read weather data from direct filepath

nicolaigoelz:
This branch / PR extends the weather module to allow loading weather data directly from a given direct_filepath and by specifying the source format. This bypasses the previous requirement to use a LocationEnum, enabling more flexible use of custom datasets.

The changes are implemented within the weather module and keep the existing LocationEnum workflow fully intact, ensuring backward compatibility.

Additionally, a minor adjustment was made to the absolute comparison in the test_generic_electrolyzer_h2.py test to avoid numerical errors.

### DIFF
--- a/.github/workflows/codequality.yml
+++ b/.github/workflows/codequality.yml
@@ -21,14 +21,15 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install prospector[with_everything]
-        pip install prospector[with-everything]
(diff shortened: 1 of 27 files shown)


k-rieck (approved):
Looks good. Would you mind to write a test for the weather component (if it doesn't exist yet) and test these different input parameters  (location_entry vs path vs direct data source) to see if it all works well?

k-rieck (approved):

PR #436, opened 2026-05-18
Remove old sim repo

k-rieck:
- remove old sim_repository and use singleton_sim_repository instead
- needed to adjust tests as pytest execution failed in the CI environment
- a fixture was introduced to reset singleton instances between tests
- the simulator.run_all_timesteps() got an optional argument "clear_sim_repo" which was needed for some tests to run successfully because by default this function clears the sim repo at the end


k-rieck:
- pytest fixture removed as not needed anymore
- SingletonSimRepository.clear() just clears the dict now, it doesn't delete it anymore

noah80:
@ice2-llmbot /review

noah80:
@ice2-llmbot /improve

PR #512, opened 2026-06-24
Added 1r1c building model, refactor of building.py

Niteftef:
This PR consists of two parts:

1) I wrote a 1r1c building model. It is designed such that it can be switched with the existing 5r1c building model almost 1 to 1. I also wrote some basic tests for it. This does not interfere with the rest of hisim in any way and can therefore be safely added. Only downside could be that another component is added to the library, if the library is supposed to be kept small, then this would of course go against that.

2) I refactored the building model.

- 1) I changed the "direct_horizontal_irradiance" variable to "diffuse_horizontal_irradiance". This seems to have been a naming error. I also changed this name in the related tests, so that's why they appear in the PR.
- 2) I changed the solar gains calculation to include global irradiation. Until now, this only used the direct part of the irradiation, which seems to have been an error. All other sources I found use the global irradiation, including the RC building simulator project, which is named as a source in the module header.
- 3) On lines 1978 ff, the case indoor_air_temperature_zero_in_celsius == set_cooling_temperature_in_celsius was never handled. If that ever happened, it would have lead to an error down the line in calc_theoretical_thermal_building_demand_when_heating_or_cooling_needed_step_two().
I also have not handled it, but I included an error right there so that the problem is more easily identifiable if this ever happens.

And a bunch of stuff that only concerns the code, not the content:
- 4) Added over 60 lines of documentation
- 5) Added a lot of type annotations
- 6) Lots of small refactoring, like removing superfluous line break etc
- 7) Split up some big functions, mainly __init__() and build() to one main function and several helpers to improve readability

In total, I still shortened the entire file by over 100 lines even with the added docs.


Niteftef:
I would still add some more things to this PR after we discuss it, so probably better don't just merge before we talked about it.

Niteftef:
Oh, and to the best of my knowledge: The failing tests are not from my changes but from the existing repository.

PR #518, opened 2026-07-01
logging hotfix

Niteftef:
The AI corrections broke the logging.
The logging level is supposed to filter what gets printed to stdout,
but the AI changed it to filter everything, including what gets written
to the logfiles.
But the logfiles are supposed to contain all logs.
I reverted the singular change that caused this.

### DIFF
--- a/hisim/log.py
+++ b/hisim/log.py
@@ -134,9 +134,8 @@ class Logger:
         """
         if logging_message_path is None:
             logging_message_path = self.logging_path
-        if prio > self.logging_level:
-            return
-        if not use_profile_file:
+        # if prio high enough: print to stdout
+        if not use_profile_file and prio <= self.logging_level:
             print(str(LogPrio.get_prio_string(prio)) + ":" + message)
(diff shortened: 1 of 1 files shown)


noah80:
That was not broken by accident, that was an intentional change, since the detailed logging slows down the simulation significantly and usually nobody ever looks at the log files. 

My thinking is that if you have the problem of needing actual debugging logs, just run it with the higher setting. If this is very frequently the case we could also add a seperate LogToFileLevel though to avoid spamming the console too much?

Niteftef:
Ah okay, sorry then.

Especially when running simultaneous simulations, logging to the console isn't very helpful to me, so I use logging level 0 at all times and exclusively use the logfiles. Implementing two seperate logging levels makes things more complicated again and I'm not sure if it's worth it for such a small issue.

Another option for someone who doesn't want the console output is also rerouting stdout (I believe Jonas has done this in the past), so alternatives already exist. I will simply close this PR request.

PR #523, opened 2026-07-09
New Cost Module - First Draft

noah80:
# `hisim.economics` — the lifecycle cost engine

This package is the parallel successor of HiSim's per-component capex/opex cost calculation.
It computes **lifecycle costs over a configurable horizon** (default 20 years, annuity method
per VDI 2067-1 / DIN EN 15459-1) from three kinds of inputs: *facts* declared by components
(what am I, how big), *energy flows* measured by the meters (what crossed the system boundary),
and *versioned data files* (prices, lifetimes, subsidy schemes). Everything else — discounting,
replacements, subsidies, actor splits, uncertainty bands — happens in one central engine.

The full specification lives in **`cost_spec.md`** at the repo root; every provisional decision
and every legacy bug found during implementation is tracked in **`cost_module_issues.md`**.


nicolaigoelz:
Silent data loss on duplicate (component_type, valid_from_year) + no versioning for revised future forecasts

Two related points in database.py:

1. The loader doesn't enforce the "one entry per (component_type, valid_from_year)" rule.

_load_devices appends every entry without a uniqueness check, so two entries with the same (component_type, valid_from_year) both load without error. get_device_entry then returns max(candidates, key=lambda e: e.valid_from_year), which on a tie returns whichever entry comes first in the file — the second becomes dead data, silently, with no warning. Same pattern in _load_energy_prices / get_energy_price. The README/spec state this rule as a constraint, but the code treats it as a convention it doesn't enforce.
Suggested fix (independent of point 2): raise CostDataError on a duplicate key at load time, so a duplicate entry fails loudly instead of dropping an entry silently.


2. No clean way to revise a forward-looking price forecast.

The valid_from_year mechanism handles historical prices well (add a new entry, never edit an old one → old basis years stay reproducible). But it can't express a revised future forecast: if a valid_from_year: 2030 entry is added in 2024 and needs revising in 2026, overwriting it breaks reproducibility of the 2024 study, and a second 2030 entry doesn't help (see point 1 — the lookup can't distinguish them). Root cause: the schema has one dimension (target year) where a forecast needs two — target year and forecast vintage.

Possible solution:
add an optional forecast_vintage field so (component_type, valid_from_year, forecast_vintage) is unique and studies can pin a vintage

Checked against database.py and validation.py: the duplicate isn't caught at load time or by the data-file CI (validate_cost_database only checks that at least one entry exists per asset class/country, so a duplicate passes). A forecast_vintage field would likely also touch serialization/scenarios, which I haven't reviewed.

PR #545, opened 2026-08-06
Fix: KPIs "Heating/Cooling hours of heat pump" were inflated due to incorrect summation

### LINKED ISSUE #543: Fix time on heating
### Version Checks (indicate both or one)

- [x] I have confirmed this bug exists on the lastest commit on the master branch

- [ ] I have confirmed this bug exists on the lastest commit on the dev branch.

- [ ] I have confirmed this bug exists for a specific version the software.

### Issue Description

time on heating is more than 1 year

### Reproducible Example

```python
tbd
```

### Expected Behavior

kpi time on heating needs to be less than 1 year for a 1 year simulation

### pip list -v

```bash

```

nicolaigoelz:
The KPIs TimeOnHeating/TimeOnCooling are running counters of the current continuous heating/cooling streak (incremented by seconds_per_timestep on every active timestep, reset to 0 on any interruption).
The KPI calculation summed these counter values directly (sum()) instead of counting the number of active timesteps.

Fix: Count timesteps where TimeOnHeating/TimeOnCooling > 0 instead of summing the counter values, then multiply by seconds_per_timestep.


k-rieck (approved):

nicolaigoelz:
Waiting for approval of PR #554, after which testing should be working again

noah80:
Looks good, but please add at least one test that shows it's working correctly.

nicolaigoelz:
Testing of get_component_kpi_entries() for heating_time and cooling_time added

noah80:
Brought up to date on current main, on this branch, so the fix keeps your authorship:

- Your fix for `more_advanced_heat_pump_hplib.py` and its regression test are unchanged. The test now takes the heat pump's own output objects, since `ComponentOutput` requires a component identity these days; the assertions are as you wrote them.
- The `advanced_heat_pump_hplib` half is dropped: that module was retired to `obsolete/` in #604 and is archived as frozen text, so it is not patched.
- The merge commit is gone; the branch is rebased.

The bug is real and was visible in the committed goldens: the heat pump sizer reported 4322 heating hours for a one-week run and 175 652 for a year. With the fix the one-week pair for `automatic_default_connections` reports 79.55 hours. The goldens for the four heat pump week pairs and one year pair are being re-blessed; the bless PR will land on this branch, then this one can merge. Thanks for the catch and the test.

PR #546, opened 2026-08-06
Move more files to obsolete 4

k-rieck:
- add singleton sim repository to obsolete


noah80:
we implemented everything in here during the energy system implementation phase 3

Issue #548, opened 2026-08-10
Inconsistent py/json scenario


valentinjanser:
### Version Checks (indicate both or one)

- [x] I have confirmed this bug exists on the lastest commit on the master branch

- [ ] I have confirmed this bug exists on the lastest commit on the dev branch.

- [ ] I have confirmed this bug exists for a specific version the software.

### Issue Description

The new scenario simple_air_conditioner_household_building_sizer has both the Python-based and the json-based variants. However, against normal procedure, the json-based variant was not generated from the Python-based variant (or vice versa), but both files were created independently. Now, they are inconsistent in form, and the consistency of their logic is unchecked.

noah80:
Resolved by #708: the v1 scenario JSONs and the translator are retired, so a Python setup no longer has an independently written JSON sibling to drift from. Its recorded twin is produced from the setup and checked by the energy-system-freshness gate.

PR #549, opened 2026-08-10
548: Make system_setup readme more clear, output test skips & make system setups consistent

valentinjanser:


noah80:
Thanks for this, and apologies for the long silence. Reviewed today against current main; here is where each of the three commits stands.

**Regenerated scenario JSON for the air-conditioner sizer** — superseded. #708 retired the v1 `*.scenario.json` files and the translator altogether. Every setup now has one recorded twin, the `energy_systems/<setup>.energy_system.yaml`, produced from the Python setup by the recorder and held current by the `energy-system-freshness` gate. The inconsistency #548 reported, a Python setup and a JSON twin written independently, no longer has a place to exist, so I am closing #548 with this PR.

**README section on which files a setup ships** — superseded by the same change. It documents the translator, `regenerate_scenario_jsons.py`, the `scenario-json-freshness` workflow and `test_json_configs.py`, all of which are gone, and recommends JSON-only scenarios, which is no longer the direction. `system_setups/README.md` was rewritten in #708 around the recorded twins.

**`pytest.ini`: `-rfE` to `-rfEs`** — still right, and worth having: a skipped test names itself and its reason in the summary block instead of being a bare count. Carried forward as a fresh two-line PR from main so it does not drag the stale branch with it: see the PR linked below.

Closing this one in favour of that PR.

PR #556, opened 2026-08-14
Cost module stack 1/8: shared kernel (uncertainty, timeline, provenance, facts, parameters)

noah80:
# PR 1/8 — Cost module stack: shared kernel

- **Branch:** `cost/01-kernel` (`eedc11ce`)
- **Base:** `main` (`c0e39eb9`)
- **Files:** 14 · **Tests:** 63 passed (`tests/test_economics_kernel.py`)

## What this is

The bottom of the lifecycle cost engine (`cost_spec.md`): the types every later slice is built
from. No cost data, no evaluator, nothing that reads a file. Everything here is verifiable by
hand.

| Module | Spec | What it holds |
|---|---|---|
| `uncertainty.py` | §3.9 | the `(min, avg, max)` band and its slot-wise arithmetic |
| `timeline.py` | §3.6 | cash-flow entries, categories, actors, NPV/pivot/liquidity views, `discount_factor()`, the sign convention |
| `provenance.py` | §3.10 | the interning ledger tying every number to a source |
| `carriers.py` | — | the energy-carrier enum |
| `facts.py` | §3.4, §9.3 | what a component declares about itself, with fail-fast validation |
| `parameters.py` | §3.2 | the economic parameter set |

Also: `cost_spec.md` (methodology), `roadmap/cost-spec-v2.md` (reviewability spec),
`cost_module_issues.md` (defect list), `hisim/economics/README.md` (package overview).

## Changes to existing files — additive only

Two files outside `hisim/economics/` are touched. Both were verified additive by diffing against
`main`; the whole file is taken, and the diff is nothing but the added lines:

- `hisim/loadtypes.py` — eight envelope-measure `ComponentType` members (+12 lines)
- `hisim/postprocessing/kpi_computation/kpi_structure.py` — optional `value_min` / `value_max` on
  `KpiEntry`, so a KPI can carry its uncertainty band (+4 lines)

## Review guide

**Read `uncertainty.py` first and slowly.** Every monetary figure in the other seven PRs is built
out of this type, so an error here is an error everywhere. Check in particular that the band
order invariant (`min <= avg <= max`) cannot be violated by any operation, and that `scale`,
`clamp_upper` and multiplication behave slot-wise.

**Then the sign convention in `timeline.py` (cost-spec-v2 W3.7).** The subtle part, and the
reason sign validation could not be switched on before: `CategoryRules.REVENUE_CATEGORIES` (which
bands get *mirrored* for slot assembly) and `CategoryRules.NEGATIVE_SIGN_CATEGORIES` (which
entries are *negative*) are deliberately **different sets**. `LOAN_DISBURSEMENT` is negative but
not mirrored;
`MODERNIZATION_LEVY` is in neither, because its sign depends on the payer — the tenant pays the
rent increase, the landlord receives it (§6.4). Confirm that reading against §3.9.

**Then `facts.py`.** This is the contract the four adopted components implement in PR 8. Ask
whether the validation is strict enough to catch a component that declares nonsense, because
nothing downstream re-checks it.

### Spec sections
§3.2, §3.4, §3.6, §3.9, §3.10, §9.3 of `cost_spec.md`; W3.7 and W4.3 of `roadmap/cost-spec-v2.md`.

### Test coverage
`tests/test_economics_kernel.py` — 63 tests, all hand-checkable:
`TestUncertainValue` (band arithmetic), `TestCashFlowTimeline` (NPV, pivots, scoping),
`TestDiscountFactor` (W4.3 — one formula for the whole module), `TestSignConvention` (W3.7 — the
payer-dependent levy, per-slot violation detection, the `validate=False` escape),
`TestProvenanceLedger`, `TestFactValidation`.

No worked examples apply yet — those arrive in PR 6 and exercise PRs 3–4.

### Review round 2 in this slice
The kernel category sets moved onto `CategoryRules`, and `BILL_CATEGORIES` is now defined here
**once** (D28 finding 14) — plausibility and views bind that object instead of each holding a
verbatim copy that could drift. Worth confirming the membership is right, since two modules now
depend on it rather than on their own list.

### Package convention introduced here
There are no module-level constants anywhere in `hisim/economics` (spec §8 **D22**). A constant
lives on the class that owns the concept (`UncertainValue.ZERO`) or on a namespace class naming
the group (`CategoryRules`, `SignExpectation`), so each taxonomy is greppable and has one
documented home. Worth checking that the grouping reads naturally — every later PR follows it.

## Push

```bash
git push origin cost/01-kernel

gh pr create --base main --head cost/01-kernel \
  --title "Cost module stack 1/8: shared kernel (uncertainty, timeline, provenance, facts, parameters)" \
  --body-file roadmap/pr_descriptions/01-kernel.md
```

### DIFF
--- a/cost_module_issues.md
+++ b/cost_module_issues.md
@@ -1,0 +1,277 @@
+# Cost module implementation — open issues and decisions to clarify
+
+Running log of ambiguities found while implementing `cost_spec.md`. Items marked **DECIDED (provisional)**
+were resolved with a documented default to keep the implementation moving; they should be reviewed.
+
+## Spec open questions (§11) — provisional defaults taken in code
+
+- **Q1 (nominal vs real rates)**: implemented nominal as documented default (interest 3 %, general escalation 2 %).
+- **Q2 (default escalation rates / CO2 paths)**: `escalation_defaults_DE.json` ships with a small,
+  clearly-labeled EXPERT_ESTIMATE table (electricity 2 %, gas 3 %, oil 3 %, others 2 %); the
+  per-asset-class investment escalation table ships **empty** per the spec proposal. CO2 price paths
+  for DE encode nEHS 2024–2026 fixed prices and an ETS2 corridor estimate from 2027 — values are
+  EXPERT_ESTIMATE and need team review against sources [33]–[35].
+- **Q4 (replacement timing)**: end-of-year convention everywhere, as proposed.
+- **Q5 (gross vs net storage)**: device entries migrated from `configuration.py` carry
+  `vat_rate: 0.19` (DE) / `0.20` (AT) and are stored **as-is** from the legacy dicts. The legacy
+  numbers' VAT status is *undocumented* — parity requires using them unchanged, so the migrated
+  entries are flagged `"price_basis": "AS_LEGACY"` and the gross-up for FINANCIAL accounting is a
+  no-op for them. Needs a data-review PR to reclassify each entry as net and set true VAT handling.
+- **Q6 (anyway-cost threshold)**: default 2 years, credit on by default, gross figures always reported.
+- **Q10 (EEG feed-in)**: fixed nominal for 20 years from install (escalation 0.0 default).
+- **Q14 (strictness rollout)**: `strict_cost_completeness` defaults to warning; tests/CI use strict=True.
+- **Q15 (capacity-field convention)**: dataclass field metadata `field(metadata={"capacity": True})`.
+- **Q16 (spot series)**: no licensed EPEX data shipped; a synthetic reference profile generator is
+  provided for tests plus a documented CSV loader for user-supplied series.
+- **Q17 (spread escalation)**: defaults to the carrier escalation rate.
+- **Q18 (default counterfactual)**: tariff counterfactual is always computed when a DYNAMIC contract is
+  active; behavioral counterfactual is a second run (not automated).
+- **Q19 (§14a depth)**: grid-fee discount as tariff data only; no dimming simulation requirement.
+- **Q21 (cube explosion)**: warn > 1 000 scenarios, error > 100 000.
+- **Q25 (bands beyond money)**: service lives and emission factors stay exact in v1.
+- **Q27 (source granularity)**: per-entry `source_ids` mandatory, per-field `field_sources` optional.
+- **Q28 (registry scope)**: one `sources.json` per directory (cost_database and subsidy_catalog).
+- **Q30 (subsidy overlays)**: data overlays on subsidy catalog entries are accepted by the schema and
+  re-run scheme selection per scenario.
+
+## New issues found during implementation
+
+1. **Legacy capex dict has meter "maintenance rates" > 1** (`ELECTRICITY_METER: 2.4`,
+   `GAS_METER: 1.8` in DE/2024, i.e. 240 %/180 % of investment per year — these encode absolute
+   yearly fees, not rates). Migrated 1:1 into `devices_DE.json` for parity, but they violate the
+   sanity range a maintenance *rate* should have. The new schema has
+   `fixed_operation_cost_in_euro_per_year` for exactly this; a data PR should move
+   240 €/a resp. 360 €/a there and zero the rate — that would be a deliberate, visible KPI delta.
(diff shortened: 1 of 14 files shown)


k-rieck (approved):
Approve but a few questions or comments:
- in uncertainty.py we should replace min/avg/max by min/base/max because we do not calculate any average value, it is just a guess. therefore average value could be misleading
- in carriers.py I'd suggest replacing "hydrogen" by "green hydrogen" to make this more explicit
- I do not understand yet why timeline.py and parameters.py both have the function "discount_factor". is there an overlap of these two modules?
- in loadtypes.py I would include a few more renovation measures. according to our current renovation measure database we have this list of building elements and refurbishment measures: {external wall: [external insulation, internal insulation], roof: [insulation between joints, insulation over joints, warm roof insulation], top floor ceiling: [upper insulation], basement ceiling: [bottom insulation], basement floor and walls: [walls internal insulation], ground-contacting components: [external ground insulation], windows: [triple glazed]}

noah80:
Thanks for the review! All four points are addressed; the branch (and the whole stack above it,
since two of the changes ripple into later slices) has been updated.

**1. min/avg/max naming — agreed, renamed to `best_estimate`.** "avg" did suggest a computed
average. We went with `best_estimate` rather than `base` because "base" reads ambiguously next
to the baseline-scenario vocabulary elsewhere in the module, and "best estimate" is the
established actuarial term and self-explaining in reports. The rename covers all three layers so
the vocabulary cannot drift apart: the `UncertainValue.best_estimate` field and the
`Slot.BEST_ESTIMATE` evaluation world, the JSON band keys
(`{"min": .., "best_estimate": .., "max": ..}` in the data files and every export), and the CSV
column names. Spec and docstring prose were updated where "average" named the slot; genuine
statistical averages (e.g. the dynamic-tariff decomposition against the year's average price)
keep their wording. Report goldens regenerated.

**2. hydrogen → green hydrogen — kept `HYDROGEN`, deliberately.** The carrier enum names the
commodity bought at the system boundary; whether it is green is a property of the specific
procurement — the price entry, its emission factor and its cited source — not of the carrier
vocabulary. A green-hydrogen tariff is expressed by giving the `HYDROGEN` price entry a
green-supply source and emission factor; renaming the member would bake a procurement attribute
into the type and force a second member the moment a non-green price ever needs modelling.
Happy to revisit if the database should only ever carry green hydrogen.

**3. `discount_factor` in timeline.py and parameters.py — no overlap.** The formula exists
exactly once: `timeline.discount_factor` is *the* discount formula of the module (cost-spec-v2
W4.3), and every present value routes through it. `EconomicParameters.discount_factor` is a
documented one-line convenience wrapper that delegates to it — offered because most callers
already hold the parameter set and would otherwise repeat
`discount_factor(parameters.interest_rate, year)` everywhere. Its docstring states the
delegation explicitly.

**4. More renovation measures — implemented your taxonomy.** The five coarse envelope classes
are replaced by ten fine-grained measures matching the renovation measure database, element by
element: `WALL_EXTERNAL_INSULATION`, `WALL_INTERNAL_INSULATION`,
`ROOF_INSULATION_BETWEEN_JOISTS`, `ROOF_INSULATION_OVER_JOISTS`, `WARM_ROOF_INSULATION`,
`TOP_CEILING_UPPER_INSULATION`, `BASEMENT_CEILING_BOTTOM_INSULATION`,
`BASEMENT_WALLS_INTERNAL_INSULATION`, `GROUND_EXTERNAL_INSULATION` and
`WINDOWS_TRIPLE_GLAZED` (plus the existing `EXTERIOR_DOOR`, `AIR_SEALING`,
`VENTILATION_SYSTEM`). Each gets cost rows for DE and IE, 2026 and 2035 — cloned from its
element's existing template row, with per-measure cost bands and embodied CO2 (internal wall
insulation cheaper than external, over-joist roof insulation carrying the re-roofing cost,
perimeter insulation carrying the excavation). These are flagged AI estimates
(`src_ai_estimates`), like the coarse rows before them, to be replaced by reviewed data — if the
renovation measure database has reviewed cost figures for these measures, we would gladly take
them. The BEG EM subsidy schemes list all thirteen classes as eligible, with the per-element
U-value conditions mapped onto the fine classes (walls 0.20, roof and top ceiling 0.14,
basement/ground 0.25, windows 0.95).

PR #557, opened 2026-08-14
Cost module stack 2/8: cost database + shipped data files

noah80:
# PR 2/8 — Cost module stack: cost database + shipped data files

- **Branch:** `cost/02-data` (`8b1a1ae0`)
- **Base:** `cost/01-kernel` (`eedc11ce`)
- **Files:** 26 · **Tests:** 98 passed (35 new in `tests/test_economics_database.py`)

## What this is

The loader and the data it loads. The data layer imports only the kernel of PR 1,
`hisim.loadtypes`, `hisim.log` and the standard library — it has no idea an evaluator exists.

It is **three modules** (spec §8 **D21**), split by concern so the loader is readable on its own:

**`hisim/economics/catalog_entries.py`** — the row types of the cost catalog: `DeviceEntry`,
`EnergyPriceEntry`, `Co2PricePath`, `EscalationDefaults`, the resolved-entry wrappers and
`CostDataError`. Pure data with their own validation; no file I/O.

**`hisim/economics/sources.py`** — the structured source registry (§3.5): `SourceEntry`,
`SourceRegistry`, its mandatory-field rules and the orphan bookkeeping that `resolve()` drives.

**`hisim/economics/database.py`** — the loader: device and energy-price lookup rules, CO2 price
paths, escalation defaults, provenance recording for every looked-up field (§3.10), and scenario
overlays (§4.6). It **re-exports** the row types and the registry deliberately, so the data layer
keeps one public import surface and every existing `from hisim.economics.database import …` keeps
working. `TestReExportSurface` pins that contract.

**`hisim/cost_database/`** — devices, energy prices, escalation defaults and allocation rules for
DE/AT/IE; CO2 price paths; plausibility check ranges; the default perspective bundle; the source
registry; one tariff contract; the synthetic reference spot series.

**`hisim/subsidy_catalog/`** — the DE and AT subsidy catalogs, their questionnaires and their
source registry. **Pure data in this PR** — the loader is `subsidies.py` in PR 3, so nothing
reads these files yet.

**`setup.py`** — `package_data` for the two new data directories, so an installed `hisim` ships
them alongside `hisim/inputs`.

## Review guide

**The lookup rules decide which price a simulation sees, so read them before the data.** Two
rules carry the weight:

1. *newest entry valid at the requested year wins* — check the boundary behaviour when a price
   series has a gap, and when the requested year predates every entry;
2. *missing data is a hard error*, never a silent default. Confirm there is no fallback path that
   would let a run complete with an invented number.

**Then spot-check the data against its sources.** Every datapoint must resolve to a source id in
`sources.json`; the tests enforce that mechanically, but only a human can judge whether the cited
source actually says what the number claims. `devices_DE.json` (2212 lines) and
`energy_prices_DE.json` (1578 lines) are the ones that matter most — they drive the default runs.
The AT and IE files are smaller and less exercised.

**Worth asking:** are the band widths (`min`/`max` around each `average`) defensible, or
placeholders? They set the uncertainty on every published figure, and PR 1's band arithmetic will
propagate whatever is here without comment.

## Review round 2 in this slice

**D26 — uniform EUR/kWh pricing.** `BillingDeterminants.energy_bought_in_kwh` used to hold *tons*
for pellets and wood chips and *liters* for oil, so every "EUR/kWh" label on those carriers was
wrong, and two of the eight shipped plausibility bands judged a kWh price against a per-tonne
range. The fix moves the heating value from the quantity to the price — an identity on the
product — and the **data files keep their native quotes** (EUR/t, EUR/l) because that is how the
cited sources publish them. `CostDatabase.get_energy_price` divides the working price *and* the
emission factor by the carrier's energy content and hands out a copy in EUR/kWh and kg/kWh.

The design question worth your attention: the conversion is recorded, not hidden. `EnergyContent`
names the unit divided by and where the heating value came from (`PhysicsConfig`, with diesel
falling back to the 9.8 kWh/l `generic_car` uses), so a reader of `cost_provenance.json` can redo
the division by hand rather than trust a rounded EUR/kWh figure. Check that you agree the native
quote is the right thing to ship, and that an unconvertible unit or unknown carrier failing loudly
(D7) is the right default.

**D27 — §559e data.** `allocation_DE_2024.json` gains `heating_cap_in_euro_per_m2_per_month`
(0.50) and `heating_measure_component_types`, the data-driven classification of which asset
classes count as heating modernization. Everything not on that list stays at §559's 8 %, which is
the conservative direction. The rule that consumes it is in PR 4; the hand-worked example is in
PR 6.

**On the split itself (D21).** The three modules are a pure move — no logic changed. Worth
asking whether the seam is in the right place: `catalog_entries.py` is the only module that knows
what a device row *is*, `database.py` the only one that knows where rows come *from*. Constants
follow §8 D22 and sit on their owning class: `CostDatabase.DEFAULT_PATH`,
`SourceRegistry.SOURCE_KINDS`, `DeviceEntry.PER_UNIT_TO_SIZE_UNIT`.

### Spec sections
§3.5 (source registry), §3.10 (provenance), §4.6 (scenario overlays), §5.2 (shipped data) of
`cost_spec.md`; §8 D21 and D22 of `roadmap/cost-spec-v2.md`.

### Test coverage
`tests/test_economics_database.py` — 34 tests: the shipped files all load; every datapoint
resolves a source; the lookup and overlay rules behave as specified; malformed data is
rejected (missing `sources.json`, an unsourced entry, an unknown source id, an unknown component
type); and `TestReExportSurface` asserts that `database.py` still re-exports every row type, that
the re-exports are the canonical objects rather than second definitions, and that the constants
really are class-scoped.

The catalogs added here are validated end-to-end in PR 3 (`validate_all()` reports 0 errors,
0 warnings) and priced in PR 6's worked examples.

## Push

```bash
git push origin cost/02-data

gh pr create --base cost/01-kernel --head cost/02-data \
  --title "Cost module stack 2/8: cost database + shipped data files" \
  --body-file roadmap/pr_descriptions/02-data.md
```

Retarget to `main` once PR 1 has merged.


k-rieck (commented):
A few comments and change requests before merging:

- spot series: is this synthetic reference csv AI generated? where is its source mentioned?
- cost_database/energy_prices_DE.json: some energy carriers like oil have their original opex values in euro_per_liter, in this document the same value is given as working_price_euro_per_kwh. that is wrong and can lead to confusion
- allocation_DE_2024.json: the module's docstring needs a bit more context in my opinion as it is not directly clear that this refers to the split between landlord and tenant. An example docstring could be: """Rules for allocating building-related costs between landlord and tenant.

            This module defines the legal rules used to determine which costs can be
            passed on to a tenant and which remain the landlord's responsibility. This
            includes, for example, maintenance and modernization costs as well as the
            allocation of CO2 costs associated with heating.

            The rules are jurisdiction- and year-specific and are represented as
            configuration data so that the billing engine can apply them consistently.
            """

PR #558, opened 2026-08-14
Cost module stack 3/8: subsidy engine, tariff engine, data validation

noah80:
# PR 3/8 — Cost module stack: subsidy engine, tariff engine, data validation

- **Branch:** `cost/03-subsidies-tariffs` (`8714f975`)
- **Base:** `cost/02-data` (`8b1a1ae0`)
- **Files:** 4 · **Tests:** 156 passed (58 new in `tests/test_economics_subsidies_tariffs.py`)

## What this is

Three modules, ~all of the rule complexity in the cost engine. They turn the catalogs shipped in
PR 2 into cash flows. Both engines import only the kernel (PR 1) and the database (PR 2);
neither reaches into the evaluator.

- **`subsidies.py`** (§5) — the subsidy scheme model and its solver: eligibility conditions,
  rate / lump-sum / tiered awards, caps and basis clamping, combination rules, the cumulation
  solver that picks the best legal combination, and the decision audit trail (§5.6).
- **`tariffs.py`** (§4.4) — the tariff contract model: flat, time-of-use, dynamic-spot and
  capacity-charge components, feed-in netting, controllability discounts.
- **`validation.py`** — schema and consistency validation of the shipped cost database, tariff
  contracts and subsidy catalogs.

## Review guide

This is the highest-risk diff in the stack per line. The rules encode real subsidy programmes and
real tariff structures, and a wrong rule produces a plausible-looking number that no invariant
test will catch.

**Start with the cumulation solver in `subsidies.py`.** It answers "which combination of schemes
wins", and it is the one place where a bug changes the headline result without changing anything
that looks wrong. Check it against §5: is it really choosing the best *legal* combination, and
does it reject combinations the programme rules forbid rather than silently down-weighting them?

**Then the clamping order.** Basis clamping, rate application, absolute cap and tiered cap
compose, and the order matters — a lump sum clamped to the basis before or after a cap gives
different money. §5 fixes the order; confirm the code follows it.

**Then `tariffs.py` feed-in netting.** Netting decides whether exported energy offsets imported
energy before or after the tariff is applied, which is the difference between two quite different
bills.

**Verified on this branch:** `validation.validate_all()` reports **0 errors, 0 warnings** against
all shipped data, and both catalogs load (DE: 9 schemes).

## Test coverage — 51 tests, shipped with the code they test

`tests/test_economics_subsidies_tariffs.py` — **58 tests, all passing on this branch**:

- `TestSubsidyEngine` — eligibility, award shapes, caps
- `TestTypedBenefits` — grants, tax credits, and how they differ
- `TestConditionAstAndFieldVocabulary` — the eligibility condition language
- `TestOverallCapIsAppliedPerSlot` — capping in all three worlds
- `TestSubsidyModeFiltersBeforeTheSolve` — which schemes reach the solver
- `TestCumulationSolverGuard` — the solver and the combination rules
- `TestSubsidyProvenance` — a catalog and its schemes cite their registry entries
- `TestTariffEngine` — contract evaluation, netting, capacity charges
- `TestScenarioDataOverlays` — the §4.6 overlay surface of the cost database
- `TestDeletedDeadSurface` — D23: a deleted knob stays deleted

Review round 2 adds the **D25 fail-fast** cases here: an INELIGIBLE scheme now names the
condition leaves that failed and the answers that failed them (`any` reports as "none of", `not`
says the child holds), instead of returning a bare verdict a user cannot act on. Read those
messages as a user would — they are the diagnostic surface of the whole subsidy engine.

This file previously shipped with PR 4, because nine of its tests drove the same rules *through*
the evaluator and so could not pass before that layer existed. Those nine were split out into
`tests/test_economics_subsidy_integration.py`, which ships with **PR 4**; everything that needs
nothing above the rule engines now ships here. The split was pure code motion — the nine bodies
are byte-identical and the set of test names across the two files is unchanged (69 = 58 + 11).

Reviewers who want the end-to-end picture should read the integration file in PR 4 as well: it
covers subsidy provenance reaching `explain()`, scenario-set expansion, the evaluation cube and
the break-even search.

**PR 6** is the stronger check on this diff: the `subsidies/` and `tariffs/` worked examples are
five hand-computed Excel workbooks each, covering exactly the cases above (share with absolute
cap, combined rate/cap pro-rata, lump sum clamped to basis, per-dwelling-unit tiered cap, grant
versus tax credit; flat bill, two-band ToU, dynamic spot integral, capacity charge,
controllability discount, feed-in netting).

### Spec sections
§5 and §5.6 (subsidies), §4.4 (tariffs), §9.6 (data-file CI checks) of `cost_spec.md`.

## Push

```bash
git push origin cost/03-subsidies-tariffs

gh pr create --base cost/02-data --head cost/03-subsidies-tariffs \
  --title "Cost module stack 3/8: subsidy engine, tariff engine, data validation" \
  --body-file roadmap/pr_descriptions/03-subsidies-tariffs.md
```

Retarget as the PRs below it merge.

### DIFF
--- a/cost_module_issues.md
+++ b/cost_module_issues.md
@@ -91,7 +91,7 @@ were resolved with a documented default to keep the implementation moving; they 
     BASEMENT_CEILING_BOTTOM_INSULATION, BASEMENT_WALLS_INTERNAL_INSULATION,
     GROUND_EXTERNAL_INSULATION, WINDOWS_TRIPLE_GLAZED) plus EXTERIOR_DOOR, AIR_SEALING and
     VENTILATION_SYSTEM — with AI-estimate entries for DE/IE 2026/2035, BEG EM envelope schemes
-    (15 % + 5 % iSFP, U-value conditions) plus the `building.has_isfp` question, and three
+    (15 % + 5 % iSFP, U-value conditions) plus the `building.has_renovation_roadmap` question (iSFP in the German catalog; renamed 2026-08-18 per PR-3 review), and three
     engine mechanics (see README §3.2b).
     Decisions taken:
     - **(a)** `energy_related_cost_share` implemented as the coupled-cost (Ohnehin-Kosten)
--- a/hisim/economics/README.md
+++ b/hisim/economics/README.md
@@ -287,7 +287,7 @@ With 35–50 a service lives inside a 20 a horizon, envelope economics are domin
 **residual value** (half the wall insulation cost is credited back at year 20 of a 40 a life)
 — that is the intended VDI 2067 treatment, not a bug. The DE catalog contains the BEG EM
 envelope schemes (15 % base with per-class U-value conditions + 5 % iSFP bonus, capped at
-20 %, mutually exclusive with §35c) and the `building.has_isfp` questionnaire entry.
+20 %, mutually exclusive with §35c) and the `building.has_renovation_roadmap` questionnaire entry (iSFP in the German catalog).
 
 `per_unit` is one of `"kW"`, `"kWh"`, `"liter"`, `"m2"` or `null` (absolute price per device);
 it must match the `size_unit` the component declares — the pre-run resolution check enforces
--- a/hisim/economics/subsidies/__init__.py
+++ b/hisim/economics/subsidies/__init__.py
@@ -1,0 +1,128 @@
+"""Data-driven subsidy engine: EU scheme modeling (cost_spec.md §5).
+
+Schemes live in ``hisim/subsidy_catalog/<COUNTRY>.json``. Eligibility is a small data-only
+predicate language over a typed context; unanswered questions yield tri-state eligibility
+(§5.7). The cumulation solver enumerates admissible combinations and picks the
+NPV-maximizing one on the BEST_ESTIMATE slot, then values it in all three slots (§5.4).
+
+This module answers one question — *how much support does this measure get, from which schemes,
+and on what evidence* — and answers it entirely from data, so that adding a country or a funding
+programme is a catalog edit rather than a code change (§10.1: if the engine needs changing for a
+new country, the schema failed). It is split into two halves by the ``=== engine`` banner in the
+middle of the file: above it the inert catalog payload (schemes, benefits, conditions as data,
+the loader), below it everything that reads a *context* and decides anything (condition
+evaluation, eligibility assessment, cumulation solving, question derivation). The cut is
+deliberate and prepares the §2.5 package split into ``economics/data/`` and ``economics/engine/``.
+
+What the module deliberately does NOT own: cash flows, timelines, discounting policy, perspective
+types and VAT netting. It returns a :class:`SubsidyDecision` of abstract awards; turning those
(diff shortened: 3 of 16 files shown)


k-rieck (commented):
- in some subsidy classes like subsidybuildingcontext some parameters have only a German context like "has_ispf" or "WEG" in Applicant Actor. Why not make this more general and applicable to other European countries by naming it "renovation_roadmap" or "condiminoum association"? see also https://www.mdpi.com/2071-1050/17/5/2289 for other countries "ispfs"
- subsidy module has 218 lines. that is too long in my opinion. is it possible to split it to get an easier overview?
- can test_economics_subsidies-and_tariffs be separated into two tests? one testing subsidies, one testing tariffs? this test has 1082 lines ...
-

noah80:
**1. German-only context parameters — generalized.** You are right that the shared vocabulary
should not bake German instruments in. `SubsidyBuildingContext.has_isfp` is now
`has_renovation_roadmap` — individual renovation roadmaps are an EU-wide instrument (the survey
you linked, Sustainability 17(5):2289, is now cited at the field) and the iSFP is the German
instance, which keeps its name in the German questionnaire labels and the `..._ISFP_...` scheme
id. `ApplicantActor.WEG` is now `CONDOMINIUM_ASSOCIATION`, with "Wohnungseigentümergemeinschaft
(WEG)" remaining as the German label in `questions_DE.json`. The dividing line we applied: the
*context vocabulary* (what conditions and questionnaires speak) is country-neutral; the country
*catalogs* stay free to be as German (or Austrian, or Irish) as the law they encode.

**2. subsidies module too long — split into a package.** `subsidies.py` (2 191 lines, of which
1 013 are code — the rest is the module's documentation) is now the package
`hisim/economics/subsidies/` with four modules cut along its natural seams, each under 500 code
lines: `context.py` (applicant/building context and the field vocabulary), `catalog.py`
(conditions, benefits, schemes, catalog loading), `assessment.py` (condition evaluation and
scheme assessment), `solver.py` (award math and the cumulation solver). Pure code motion — the
package `__init__` re-exports the entire previous surface, so no import anywhere changed.

**3. Test file split — done, and applied as a general rule.** `test_economics_subsidies_tariffs.py`
is now `test_economics_subsidies.py`, `test_economics_subsidy_solver.py` and
`test_economics_tariffs.py`. We then checked the whole stack (PR 3–9) against a ~500-code-line
rule (docstrings and comments not counted, since the module documentation is deliberately heavy):
the only other production module over the limit was `reporting.py` (1 298 code lines), which is
now the package `hisim/economics/reporting/` (summary / charts / sections / assembly), and six
test files were over, all now split by theme with cross-referencing docstrings:
the engine tests into three files, the data/integration tests into three, the property-test
(invariants) file into four, the reporting tests into two, and the worked-example group runners
moved into a non-test library module (`tests/worked_example_runners.py`). Every economics
production and test file is now under 500 code lines, and the import-lint learned to treat a
package as one unit so the presentation/engine seam rules apply unchanged.

k-rieck (approved):

PR #559, opened 2026-08-14
Cost module stack 4/8: the evaluation engine

noah80:
# PR 4/8 — Cost module stack: the evaluation engine

- **Branch:** `cost/04-engine` (`26f69158`)
- **Base:** `main` (`60a69e6a`, after the 3/8 merge)
- **Files:** 34 · **Tests:** 315 passed (156 new here)

## Rebase note (2026-08-21)

PR 3/8 is merged, so this PR now targets `main` directly. The branch was rebased onto the
post-merge `main` (clean — this slice only adds files) and gained one commit, `26f69158`,
fixing the same quality-pipeline error classes that the 3/8 merge surfaced on `main`, so this
PR's pipeline is green before merging. Almost all of it is mechanical (mypy annotations,
pylint/prospector style), with **one real fix a reviewer should look at**:
`LifecycleCostResult.to_json` called `EconomicParameters.to_dict()`, which did not exist
anywhere in the stack — the method now exists on `EconomicParameters` (plain `asdict`; every
enum in the record is a str-enum, so the dict stays JSON-serializable).

## What this is

The engine proper: everything that turns facts + database + rules into a cash-flow timeline and
the result objects built on it.

| Module | What it does |
|---|---|
| `evaluator.py` | the single entry point (§3.1): assembles the timeline for one variant under one perspective, in three coherent worlds |
| `calculators/` | the per-concern calculators the evaluator drives — investment, replacement/residual value, maintenance, energy, escalation, CO2, subsidy application, financing application, reserve, context resolution, category sets, aggregation |
| `financing.py` | loan schedules: annuity, interest-only bullet, soft loans, partial shares |
| `perspectives.py` | the perspective model (§4.2) and default bundle selection |
| `actors.py` | actor allocation rulesets (§6) |
| `results.py` | `LifecycleCostResult` / `EvaluationMatrix` and the variant comparison math |
| `views.py` | the view model (W4.1): every derived number a report may show, computed once, here |
| `plausibility.py` | the typed plausibility findings (§7.4) |
| `scenarios.py` | the scenario cube, tornado data, break-even search (§4.6) |

`requirements.txt` gains **`numpy_financial`**, used only by the invariant tests as an
independent cross-check of the closed-form financial math.

## Review round 2 corrections in this slice

Each is pinned by a test that fails without it (spec §8 D23/D24/D27):

- **Residual value is gated on an installation the timeline actually charged.** Per
  DIN EN 15459-1, a kept brownfield asset outliving the horizon no longer credits revenue against
  an investment nobody paid. This one changes headline numbers — check it first.
- **`operational_co2_by_carrier_in_kg` accumulates instead of assigning**, so two meters on one
  carrier agree with the by-year series.
- **Loan-award substitution uses `is not None`** for rate, term and repayment grant, so a stated
  0.0 % soft-loan rate survives instead of being read as "unset".
- **§559e heating levy (D27)** — one deduction rule run over two pools, with the
  heating-attributable increase capped at 0.50 EUR/m²/month *nested inside* the existing §559
  Abs. 3a cap regime. The nesting order is the part to check against the statute; PR 6's
  `heating_levy_559e_mixed_package` workbook is the hand-computed cross-check.

## Review guide

**`calculators/aggregation.py` is the file to read most carefully.** Every headline KPI is
assembled there, and the per-component breakdowns must sum exactly to the totals — that is a
promise the frontends rely on. The invariant tests check it, but read the code for whether the
promise holds by construction or by coincidence.

**The three-world slot assembly in `evaluator.py`.** The engine evaluates LOW / AVERAGE / HIGH in
one pass, and the bounds are *envelopes*, not statistical quantiles. The correctness question is
whether each world is internally **coherent** — that the LOW world takes the cheap end of every
cost *and* the appropriate end of every revenue (which is where PR 1's `CategoryRules.REVENUE_CATEGORIES`
mirroring comes in). A world that mixes ends produces a band that is too narrow and indefensible.

**`views.py` and the W4.1 claim.** This PR claims that every number a report displays is derived
here. PR 7 depends on that being true — it is enforced mechanically by the import lint in PR 8.
While reviewing, ask of any derived quantity: could a report have computed this itself instead?
If yes, it belongs here.

**`actors.py` (§6).** The allocation rulesets split flows between landlord and tenant. Check the
modernization levy in particular: it is a transfer, so the same money must appear positive on one
actor and negative on the other, and the totals must still reconcile.

### Spec sections
§3.1, §4.1, §4.2, §4.6, §6, §7.4 of `cost_spec.md`; W4.1 and W4.3 of `roadmap/cost-spec-v2.md`.

### Test coverage — 307 passed (151 new here)

| File | Covers |
|---|---|
| `test_economics_engine.py` | the evaluator end to end |
| `test_economics_invariants.py` | property tests, plus the `numpy_financial` cross-check of NPV, annuity factors and loan schedules against an independent implementation |
| `test_economics_envelope.py` | the envelope-measure cost subjects (PR 1's eight `ComponentType` members) |
| `test_economics_views.py` | the view model (W4.1) |
| `test_economics_subsidy_integration.py` | the 11 **cross-layer** tests that drive PR 3's subsidy and scenario machinery *through* an engine module — a subsidy award followed from the catalog into `explain()`, the §10.1 legacy-shim provenance origin, scenario-set expansion, the evaluation cube with its tornado/swing/spread helpers, the counterfactual billing boundary, the break-even search and the D23 `FinancingPlan` dead-surface pin. The 58 unit tests of those same engines ship with PR 3. |

`test_economics_invariants.py` is the strongest automated evidence in this PR: it re-derives the
financial math with `numpy_financial` rather than restating the implementation.

The hand-computed cross-check of this engine is **PR 6** — in particular the `discounting/`,
`financing/` and `end_to_end/` worked examples.

## Push

```bash
git push origin cost/04-engine

gh pr edit 559 --base main --body-file roadmap/pr_descriptions/04-engine.md
```


hisim/economics/calculators/annualization.py
```diff
@@ -0,0 +1,166 @@
...
+
+    A single named constant rather than a literal at the division site, so the one number that
+    distinguishes the two historical guard styles (see the module docstring) is greppable and can
+    be deleted in one place once package 2b unifies them. It is a floor on the *divisor*, not a
+    physically meaningful fraction — no simulation produces a period shorter than a nanosecond of
+    a year.
+    """
+
+    #: Lower bound the subsidy call site clamps the fraction to before dividing (see module docstring).
+    ZERO_GUARD = 1e-9
```
k-rieck, line 50: that might be explained somewhere, but why do we want to have a divisor of 1e-9? that would create a non-meaningful result. allowing zeros would at least raise an error

k-rieck (approved):
please fix the tests before merging

PR #560, opened 2026-08-14
Cost module stack 5/8: serialization, component adapter, CLI

noah80:
# PR 5/8 — Cost module stack: serialization, component adapter, CLI

- **Branch:** `cost/05-io-cli` (`aa12239e`)
- **Base:** `cost/04-engine` (`d0bf105f`)
- **Files:** 5 · **Tests:** 307 passed (no *new* test file here — see below)

## What this is

Everything that moves engine results across a process boundary. No new economics behaviour —
this slice is about file formats.

- **`serialization.py`** (§3.8) — the JSON round trip of the evaluator's inputs and results, so a
  run can be re-priced later without re-simulating.
- **`exports.py`** — the result files themselves (`lifecycle_costs.json`,
  `cash_flow_timeline.csv`) and the lifecycle KPI entries, so the KPI table and
  `lifecycle_kpis.json` cannot disagree.
- **`input_audit.py`** (W4.6) — the typed "which input came from where" rows. The module that
  *fills* them from the cost database is `audit.py` in PR 7.
- **`adapter.py`** (§3.4) — maps a HiSim component's config to `ComponentCostFacts`. Read-only
  towards the components; the components themselves are not touched until PR 8.
- **`__main__.py`** — the `python -m hisim.economics` CLI: `evaluate` / `explain` / `report` /
  `validate`.

## Placement note

`exports.py` and `input_audit.py` were originally grouped with the presentation PR, but
`serialization.py` imports `exports` at module level and `__main__.py` imports `input_audit` at
module level — so they sit *below* the presentation layer, not in it. They are byte-identical to
`new_cost_module`; only their PR assignment moved.

## Known gap in this slice

The CLI's `report` and `explain --audit` paths import `reporting` / `report_plots` / `audit`
lazily, inside the subcommand handlers. Those modules arrive in **PR 7**, so on this branch:

- `python -m hisim.economics validate` works — **verified: 0 errors, 0 warnings**
- `python -m hisim.economics evaluate` works
- `python -m hisim.economics report` raises `ImportError`

That is the existing lazy-import structure of the file (the usual pattern for deferring
matplotlib), not something introduced by the split. It closes in PR 7.

## Review guide

**The JSON schemas in `serialization.py` are the compatibility surface of this whole module.**
Once results are stored in the field, changing these shapes breaks the ability to re-price old
runs. Review them as you would a public API: are the field names ones you want to live with, are
optional fields genuinely optional, and does the round trip lose anything (bands, provenance ids,
subsidy decision trails)?

**`exports.py` and the KPI consistency claim.** The reason `build_lifecycle_kpi_entries` lives
here rather than in the report is so the KPI table and `lifecycle_kpis.json` are produced from one
source. Check that PR 7's report really consumes this and does not build a parallel table.

**`adapter.py`** is the seam between HiSim components and the cost engine. Ask whether it fails
loudly on a component it does not understand, or silently returns nothing — the latter would mean
a component silently priced at zero.

### Spec sections
§3.4 (facts), §3.8 (serialization), §3.10 (provenance) of `cost_spec.md`; W4.6 of
`roadmap/cost-spec-v2.md`.

### Test coverage — read it in PR 8

**This PR ships no test file of its own.** `tests/test_economics_data_and_integration.py` covers
serialization, exports and the CLI (49 tests), but it also asserts the cost-fact contract of the
four components adopted in PR 8 and calls the parity harness in `audit.py` from PR 7 — so it can
only pass once both have landed, and is committed with **PR 8**.

The classes covering this diff are `TestSerializationRoundtrip` (round trips, contracts, subsidy
context) and `TestCliAndExports` (the CLI and the export files). The 307 tests of PRs 1–4 still
pass on this branch.

## Push

```bash
git push origin cost/05-io-cli

gh pr create --base cost/04-engine --head cost/05-io-cli \
  --title "Cost module stack 5/8: serialization, component adapter, CLI" \
  --body-file roadmap/pr_descriptions/05-io-cli.md
```

### DIFF
--- a/hisim/economics/serialization.py
+++ b/hisim/economics/serialization.py
@@ -1,0 +1,730 @@
+"""Serialization of evaluator inputs for post-hoc re-pricing (cost_spec.md §4.6).
+
+All evaluator inputs are serialized into the result directory (`economic_inputs.json`,
+accompanied by `cost_provenance.json`) so new economic assumptions never require re-running
+the building simulation.
+
+**This module implements the seam-1 contract of cost-spec-v2 §2.1.** The engine is required to be
+a pure function of a JSON file: everything the evaluator needs about one simulated variant must
+survive being written to disk and read back, so that re-pricing an archived run under new interest
+rates, an updated subsidy catalog or a reviewer's "what if" never re-runs the building simulation.
+That is also what makes the two error classes separable — "the wrong physical quantities went in"
+is a question about this file, "they were priced wrong" is a question about everything downstream
+of it — and what lets the downstream tests be small hand-written JSON files with exact expected
+results instead of simulation runs.
+
+**Round-trip fidelity is therefore the property under test**, not an implementation detail:
+`tests/test_economics_data_and_integration.py` writes inputs, reads them back and asserts that the
+reloaded record evaluates to the same NPV per slot, that a `SubsidyBuildingContext.existing_heating`
+survives with every attribute (it decides the BEG speed bonus — dropping it used to silently degrade
+that scheme, cost-spec-v2 W1.3), and that a `TariffContract` which exists in no catalog file
+round-trips byte-equal and prices identically (W1.4). A field that does not round-trip is a bug of
+exactly the class this seam exists to prevent: it makes an archived run re-price differently from
+the run that produced it, silently and without any error to catch. The one *deliberate* exception is
+the default tariff contract synthesized from the §3.5 price entries — it is written for the record
+but not restored, because it is derived data the evaluator regenerates at the price basis year, and
+restoring it would freeze the shipped prices and defeat scenario price overlays.
+
+The module owns nothing economic: no defaults, no fallbacks, no validation beyond what the target
+types enforce in their own constructors. It also owns the *inverse* direction (below the "stored
+results" divider, W4.5): reading `lifecycle_costs.json`, `cash_flow_timeline.csv` and
+`cost_provenance.json` back into a full `EvaluationMatrix`, so `python -m hisim.economics report`
+renders a stored evaluation instead of quietly performing a new one.
+"""
+
+from __future__ import annotations
+
+import csv
+import json
+import os
+from typing import Any, Dict, Optional
+
+from hisim.economics.carriers import EnergyCarrier
+from hisim.economics.evaluator import EvaluationInputs, SubjectCostFacts, UnresolvedSubject
+from hisim.economics.exports import ExportFileNames
+from hisim.economics.facts import (
+    BillingDeterminants,
+    ComponentCostFacts,
+    ExistingAsset,
+    ExistingAssetRegister,
+)
+from hisim.economics.parameters import EconomicParameters
+from hisim.economics.provenance import ProvenanceLedger
+from hisim.economics.results import (
+    AnnualEnergyQuantities,
+    ComponentCostBreakdown,
+    EvaluationMatrix,
+    LifecycleCo2Result,
+    LifecycleCostResult,
+    ReferenceAreas,
+)
+from hisim.economics.subsidies import (
+    ApplicantActor,
+    ApplicantProfile,
+    HeritageStatus,
+    PayoutKind,
+    SubsidyAward,
+    SubsidyBuildingContext,
+    SubsidyContext,
+    SubsidyDecision,
+)
+from hisim.economics.tariffs import contract_to_json
+from hisim.economics.timeline import Actor, CashFlowEntry, CashFlowTimeline, CostCategory, SubjectKind
+from hisim.economics.uncertainty import UncertainValue
+from hisim.loadtypes import ComponentType, Units
+from hisim.postprocessing.kpi_computation.kpi_structure import KpiTagEnumClass
+
+
+class SerializationFileNames:
+    """Names of the files the input/provenance serialization writes.
+
+    Kept as named constants rather than literals because both directions of the seam and several
+    unrelated readers (the CLI, the parity pass, the report layer, archived-result tooling) address
+    the same two files; the export-side names live in `exports.ExportFileNames`.
+    """
+
+    ECONOMIC_INPUTS_FILE_NAME = "economic_inputs.json"
+    PROVENANCE_FILE_NAME = "cost_provenance.json"
+
+
+def _band_or_none(value: Optional[UncertainValue]) -> Any:
+    """Serializes an optional uncertainty band, preserving the None/zero distinction.
+
+    Every monetary override is optional, and "no override" is a different statement from "an
+    override of zero euro" — so None must survive as JSON null rather than collapsing to a band.
+    """
+    return value.to_json() if value is not None else None
+
+
+def _band_from(value: Any) -> Optional[UncertainValue]:
+    """The inverse of `_band_or_none`: JSON null stays None, anything else becomes a triplet."""
+    return UncertainValue.from_json(value) if value is not None else None
+
+
+def facts_to_json(facts: ComponentCostFacts) -> dict:
+    """Serializes ComponentCostFacts.
+
+    Writes every declared field including the per-field overrides and `override_source`, so a
+    re-priced run applies exactly the same overrides — and stays as traceable — as the original.
+    Enums are stored by `name` (asset class, size unit) and the KPI tag by its value; the free-form
+    `technical_attributes` dict passes through unchanged, which is why `ComponentCostFacts` requires
+    it to be JSON-serializable in the first place.
+    """
+    return {
+        "asset_class": facts.asset_class.name,
+        "size": facts.size,
+        "size_unit": facts.size_unit.name,
+        "kpi_tag": facts.kpi_tag.value if facts.kpi_tag else None,
+        "count": facts.count,
+        "investment_cost_override_in_euro": _band_or_none(facts.investment_cost_override_in_euro),
+        "installation_cost_override_in_euro": _band_or_none(facts.installation_cost_override_in_euro),
+        "lifetime_override_in_years": facts.lifetime_override_in_years,
+        "maintenance_rate_override": _band_or_none(facts.maintenance_rate_override),
+        "fixed_operation_cost_override_in_euro_per_year": _band_or_none(
+            facts.fixed_operation_cost_override_in_euro_per_year
+        ),
+        "embodied_co2_override_in_kg": facts.embodied_co2_override_in_kg,
+        "override_source": facts.override_source,
+        "technical_attributes": facts.technical_attributes,
+    }
+
+
+def facts_from_json(raw: dict) -> ComponentCostFacts:
+    """Deserializes ComponentCostFacts.
+
+    The exact inverse of `facts_to_json`. Optional keys are read with `.get` and their dataclass
+    defaults, so a file written by an older version still loads; note that this reconstruction runs
+    through `ComponentCostFacts.__post_init__`, i.e. the same fail-fast validation a component
+    declaration goes through — a hand-edited inputs file with an implausible size or an unpriceable
+    size unit is rejected here rather than producing nonsense downstream.
+    """
+    from hisim.postprocessing.kpi_computation.kpi_structure import KpiTagEnumClass
+
+    kpi_tag = None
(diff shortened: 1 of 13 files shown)


hisim/economics/serialization.py
```diff
@@ -0,0 +1,728 @@
...
+        "subsidy_context": subsidy_context_to_json(inputs.subsidy_context),
+        # Contracts are embedded in full (W1.4): the file is self-contained, so re-pricing uses
+        # byte-identical contract data instead of whatever the tariffs directory happens to hold.
+        "tariff_contracts": {
+            carrier.value: contract_to_json(contract) for carrier, contract in inputs.tariff_contracts.items()
+        },
+        "consumed_tariff_ids": inputs.consumed_tariff_ids,
+        "annual_heat_demand_in_kwh": inputs.annual_heat_demand_in_kwh,
+        "building_specific_emissions_in_kg_per_m2_a": inputs.building_specific_emissions_in_kg_per_m2_a,
+        "heated_floor_area_in_m2": inputs.heated_floor_area_in_m2,
```
k-rieck, line 364: do we need both inputs? In Tabula it was stated that A_C_ref = 1.1 * A_living (so conditioned floor area = 1.1 * living area)

k-rieck (approved):

hisim/economics/serialization.py
noah80: i think yes, because tabula is nothing but an estimate for residential buildings. people could have completely different values in real houses.

noah80:
Amendment after the aggregated LLM review of this PR (and an owner decision it triggered): the cost adapter and the stack above it lose the ability to silently drop a component from lifecycle costing.

## What the review found (verified)

- The `BY_CLASS_NAME` table keyed district and electric heating as `GenericDistrictHeating`/`GenericElectricHeating` — names no class ever had — and read the nonexistent `connected_load_w`; `ModularHeatPump` and `HeatDistributionSystem` matched nothing either. Affected components fell to `UNDECLARED` and were silently never priced.
- `_hds_facts` matched the emitter against IntEnum-era ordinals and title-case literals; against today's string enum it matched nothing, so every radiator/floor-heating setup aborted the evaluation as an unresolved subject.
- An extractor tripping over a renamed config field was downgraded to a log warning and an empty extraction — indistinguishable from an unknown class.

## Owner decision: fail loudly, permanently

Any undeclared component crashes the calculation; all components are forced to declare.

**On this branch (`cost/05-io-cli`):**
- Table keys and field fixed; dead entries deleted; `_hds_facts` matches enum member names (`FLOORHEATING`/`RADIATOR`).
- Every no-facts path now carries an `unresolved_reason` (D7-fatal downstream): the extractor-accident branch, and a class declared `PRICED` that nothing can describe.
- `effective_cost_relevance` no longer infers `PRICED`/`METER` from the adapter's own tables — the class declaration is the only source of truth.
- `tests/test_economics_adapter_contract.py` is the contract test the docstrings promised: every table key must resolve to exactly one real class in `hisim.components`, and every extractor runs against every real default config — a renamed class, config field or enum now fails CI instead of quietly shrinking a cost report.
- `cost_spec.md`: the never-implemented `strict_cost_completeness` warning mode is struck (open question 14 decided: no flag, no leniency).

**On the rebased stack above (visible when 6/8–8/8 are pushed):**
- All 77 component classes declare `cost_relevance` in their own class body (30 PRICED / 4 METER / 43 FREE_OF_COST). Real hardware without a database row yet (electrolyzers, CHP, transformer, wind turbine, cars, H2 storage, air conditioners, …) is declared `PRICED` and loudly fails any lifecycle-cost run containing it until it gets a row — never `FREE_OF_COST`, which would be a lie.
- The bridge turns an `UNDECLARED` component into an `UnresolvedSubject` (D7 abort) instead of an INFO-level skip, and postprocessing re-raises `CostDataError`/`UnresolvableSubjectsError` instead of swallowing them: a run that asked for lifecycle costs and cannot have them fails.
- `Simulator.check_cost_declarations()` refuses the run **before the first timestep** (spec §9.2 "at simulation start, not end"), so a year-long simulation cannot run for hours and then die in postprocessing.
- An unreadable cost database also fails the run now instead of logging and skipping the cost files.
- Fleet contract test: any new `Component` subclass without an explicit declaration fails CI.

Note for reviewers of shipped examples: ten table-backed components (both storages, `HeatDistribution`, `MoreAdvancedHeatPumpHPLib`, EMS, …) were previously silently unpriced; lifecycle-cost reports will legitimately grow.

noah80:
Amendment after the aggregated LLM review of this PR (and an owner decision it triggered): the cost adapter and the stack above it lose the ability to silently drop a component from lifecycle costing.

## What the review found (verified)

- The `BY_CLASS_NAME` table keyed district and electric heating as `GenericDistrictHeating`/`GenericElectricHeating` — names no class ever had — and read the nonexistent `connected_load_w`; `ModularHeatPump` and `HeatDistributionSystem` matched nothing either. Affected components fell to `UNDECLARED` and were silently never priced.
- `_hds_facts` matched the emitter against IntEnum-era ordinals and title-case literals; against today's string enum it matched nothing, so every radiator/floor-heating setup aborted the evaluation as an unresolved subject.
- An extractor tripping over a renamed config field was downgraded to a log warning and an empty extraction — indistinguishable from an unknown class.

## Owner decision: fail loudly, permanently

Any undeclared component crashes the calculation; all components are forced to declare.

**On this branch (`cost/05-io-cli`):**
- Table keys and field fixed; dead entries deleted; `_hds_facts` matches enum member names (`FLOORHEATING`/`RADIATOR`).
- Every no-facts path now carries an `unresolved_reason` (D7-fatal downstream): the extractor-accident branch, and a class declared `PRICED` that nothing can describe.
- `effective_cost_relevance` no longer infers `PRICED`/`METER` from the adapter's own tables — the class declaration is the only source of truth.
- `tests/test_economics_adapter_contract.py` is the contract test the docstrings promised: every table key must resolve to exactly one real class in `hisim.components`, and every extractor runs against every real default config — a renamed class, config field or enum now fails CI instead of quietly shrinking a cost report.
- `cost_spec.md`: the never-implemented `strict_cost_completeness` warning mode is struck (open question 14 decided: no flag, no leniency).

**On the rebased stack above (visible when 6/8–8/8 are pushed):**
- All 77 component classes declare `cost_relevance` in their own class body (30 PRICED / 4 METER / 43 FREE_OF_COST). Real hardware without a database row yet (electrolyzers, CHP, transformer, wind turbine, cars, H2 storage, air conditioners, …) is declared `PRICED` and loudly fails any lifecycle-cost run containing it until it gets a row — never `FREE_OF_COST`, which would be a lie.
- The bridge turns an `UNDECLARED` component into an `UnresolvedSubject` (D7 abort) instead of an INFO-level skip, and postprocessing re-raises `CostDataError`/`UnresolvableSubjectsError` instead of swallowing them: a run that asked for lifecycle costs and cannot have them fails.
- `Simulator.check_cost_declarations()` refuses the run **before the first timestep** (spec §9.2 "at simulation start, not end"), so a year-long simulation cannot run for hours and then die in postprocessing.
- An unreadable cost database also fails the run now instead of logging and skipping the cost files.
- Fleet contract test: any new `Component` subclass without an explicit declaration fails CI.

Note for reviewers of shipped examples: ten table-backed components (both storages, `HeatDistribution`, `MoreAdvancedHeatPumpHPLib`, EMS, …) were previously silently unpriced; lifecycle-cost reports will legitimately grow.

noah80:
Second amendment round: the remaining review findings, all approved by the owner.

**CLI consolidation (`__main__.py`)** — the evaluator-setup block that was pasted three times now lives once (`_build_context` / `_evaluate_perspectives` / `_write_scenario_cube`), which fixes the drift the reviewers caught at its root:
- `--parameters` and `--subsidy-catalog` work on `evaluate`, `report` and `explain` alike (flag > parameters-file path, decided in one place).
- `report` with a re-pricing flag **re-evaluates and says so** instead of silently rendering the stored numbers under the original assumptions; `--compare` goes through the same path, so both sides of a comparison are priced under the same assumptions.
- `explain` loads the subsidy catalog, so it traces the numbers the run actually published instead of the flat-shim approximation.
- `main()` keeps the docstring's promise: `OSError`/`JSONDecodeError`/`ValueError` become a one-line message and exit 2, never a traceback; genuine programming errors still propagate.
- `evaluate` probes the (until 7/8 merges) missing audit module **before** writing, so it can no longer leave a half-written export set.

**Data-layer hardening** — `EconomicParameters.from_dict` rejects a present-but-non-dict rate value with a located error (a JSON `null` used to slip through as `None`) and accepts enum member names (`"HEAT_PUMP"`) alongside values (`"HeatPump"`), with an error listing both spellings. The None-aware `Optional[UncertainValue]`↔JSON round trip lives once, as `UncertainValue.optional_to_json`/`optional_from_json` (the two private copies in `serialization.py` and `input_audit.py` are gone). `OriginKind` is a `(str, Enum)` with byte-identical values — a misspelled origin in `cost_audit.json` now fails on load instead of rendering as a blank cell.

**Default-contract recognition** — the `"_DEFAULT_"` substring probe with its hardcoded `("DE","AT")` tuple is gone. The id format moved to its owner: `TariffContract.default_contract_id()` mints it (used by `calculators/energy.py`) and `is_default_contract_id()` recognizes the full shape — which also fixes a latent bug where a synthesized contract for any third country was sent to the catalog loader and failed. A loaded contract's typed `is_default_contract` flag is additionally honored.

**Meters name their own columns** — `get_meter_spec` resolves every billing column off the meter class's own constants at runtime (`MeterOutputContracts` stores constant *names*, not column strings), keeping the no-import rule while making the class the single source. A renamed meter output now fails the extended contract test in CI; a missing constant at runtime is a `CostDataError` → D7 unresolved subject. On the wiring branch (8/8), a meter's declared column missing from the results frame — bought, sold, **or** power — aborts the evaluation instead of warning: the old behavior silently zeroed feed-in revenue and dropped capacity charges, the same defect class as an unbilled carrier.

Verification across the rebased stack: 480 economics tests green on 8/8, full base suite green except the known pre-existing `test_solar_thermal_system` failure, end-to-end `evaluate` → `report` → `report --parameters` → `explain` exercised on a real result directory, mypy clean over all 52 economics files, pylint 10.00.

noah80:
CI fix (`c76bd366`): the previous head was red on `mypy` and `pytest (base)` — and the base failure was the new adapter contract test **working as designed**. PR CI runs on the merge with `main`, and two things drifted between this branch's base and today's main:

- main retired `advanced_heat_pump_hplib` (#604), so the `"HeatPumpHplib"` table key named a class that no longer exists in `hisim.components`. Entry and its orphaned helper removed; `MoreAdvancedHeatPumpHPLib` remains the fleet's hplib heat pump.
- main's sizing kernel made `PVSystemConfig.weather_identity` sizable; the contract sweep deliberately refuses a sizable field it has never considered, so it now carries a realized identity string (the extractor never reads it).

Plus the mypy finding from CI's `tests/` pass (`mypy_moderate.ini`): the meter sweep's `uninitialized()` takes its class as `Any`, since `type.__new__` has no overload for an arbitrary class object.

Verified on the merge-with-main tree: contract + economics suites, all three CI mypy invocations, and the full base suite.

noah80:
CI fix (`c76bd366`): the previous head was red on `mypy` and `pytest (base)` — and the base failure was the new adapter contract test **working as designed**. PR CI runs on the merge with `main`, and two things drifted between this branch's base and today's main:

- main retired `advanced_heat_pump_hplib` (#604), so the `"HeatPumpHplib"` table key named a class that no longer exists in `hisim.components`. Entry and its orphaned helper removed; `MoreAdvancedHeatPumpHPLib` remains the fleet's hplib heat pump.
- main's sizing kernel made `PVSystemConfig.weather_identity` sizable; the contract sweep deliberately refuses a sizable field it has never considered, so it now carries a realized identity string (the extractor never reads it).

Plus the mypy finding from CI's `tests/` pass (`mypy_moderate.ini`): the meter sweep's `uninitialized()` takes its class as `Any`, since `type.__new__` has no overload for an arbitrary class object.

Verified on the merge-with-main tree: contract + economics suites, all three CI mypy invocations, and the full base suite.

PR #561, opened 2026-08-14
Cost module stack 6/8: hand-computed worked-example library

noah80:
# PR 6/8 — Cost module stack: hand-computed worked-example library

- **Branch:** `cost/06-worked-examples` (`ebf1a5fe`)
- **Base:** `cost/05-io-cli` (`aa12239e`)
- **Files:** 55 · **Tests:** 366 passed (59 new here)

## What this is

The independent check on the engine of PRs 3–4. Twenty-three scenarios worked out **by hand in
Excel**, converted to YAML fixtures, and replayed against the engine (cost-spec-v2 §3.3).

- **`tests/worked_examples/`** — 23 `.xlsx` workbooks with the arithmetic written out cell by
  cell, each with its generated `.yaml` fixture, across five areas:

  | Area | Examples |
  |---|---|
  | `discounting/` | annuity factor 20y@3%, NPV mixed series 4%, NPV zero-rate plain sum |
  | `financing/` | annuity loan 10y@3%, interest-only bullet 10y@4%, partial share 15y@5%, soft loan 30y@1%, zero-interest 20y |
  | `subsidies/` | share with absolute cap, combined rate/cap pro-rata, lump sum clamped to basis, per-dwelling-unit tiered cap, grant versus tax credit |
  | `tariffs/` | flat household bill, two-band ToU, dynamic spot integral, capacity charge on annual peak, controllability grid-fee discount, feed-in fixed-tariff netting |
  | `end_to_end/` | heat-pump investment and bill, PV feed-in over 10 years, replacement and residual value, **§559e heating levy on a mixed package** |

  `enforcement.yaml` records which examples carry a human review attestation.

- **`tools/convert_worked_examples.py`** — regenerates the YAML from the workbooks; `--check`
  fails if any fixture has drifted from its workbook.
- **`tests/test_worked_examples.py`** — replays every fixture against the engine, and warns per
  example when a workbook has no review attestation.
- **`tests/test_worked_examples_converter.py`** — 11 tests on the converter itself, so a fixture
  cannot drift from its workbook through a *converter* bug rather than an edited cell. The
  `--check` guarantee is only as good as the code implementing it.
- **`.github/workflows/worked-examples.yml`** — runs `--check` and the replay in CI, so a
  workbook edited without regenerating its fixture cannot merge.
- `pytest.ini` (the `worked_examples` marker), `mypy_moderate.ini`
  (`ignore_missing_imports` for `yaml`), `requirements.txt` (`pyyaml`).

**The 23rd example is new in review round 2.** `end_to_end/heating_levy_559e_mixed_package`
is the §559e heating levy of PR 4 (D27) worked out by hand over a mixed package: a heat-pump
measure levied at 10 %/year against envelope work levied at 8 %, with the heating-attributable
rent increase capped at 0.50 EUR/m²/month inside the §559 Abs. 3a regime. It is the only
independent check on that statute reading — **and like the other 22 it is unattested**, so it is
the workbook most worth a second pair of eyes in this PR.

**Verified on this branch:**
```
tools/convert_worked_examples.py --check          ->  23 worked examples: YAML matches the workbooks.
pytest tests/test_worked_examples.py              ->  48 passed
pytest tests/test_worked_examples_converter.py    ->  11 passed
```

## Review guide

**Review the workbooks, not the code.** The Python here is a fixture loader; the substance is in
the 23 spreadsheets. Each workbook is a *claim* about what the engine should produce for a case
somebody worked out independently. If a workbook is wrong, the engine is wrong in exactly that
way and **no other test in this stack will catch it** — the invariant tests in PR 4 check internal
consistency, not external truth.

Suggested order, cheapest signal first:

1. **`discounting/`** — three workbooks, arithmetic anyone can verify in a minute. If these are
   right, PR 4's discounting is right.
2. **`financing/`** — five loan shapes. Check the zero-interest and interest-only cases, which are
   the edge cases where a closed-form implementation usually goes wrong.
3. **`subsidies/`** — five workbooks, and the strongest available check on PR 3's cumulation
   solver and clamping order. `combined_rate_cap_pro_rata` and `lump_sum_clamped_to_basis` are the
   two that pin down the ordering questions raised in PR 3's review guide.
4. **`tariffs/`** — six workbooks; `feed_in_fixed_tariff_netting` settles the netting question
   from PR 3.
5. **`end_to_end/`** — three workbooks that exercise the whole chain including replacement and
   residual value.

**The 22 attestation warnings are the point of `enforcement.yaml`,** not noise: they are the list
of workbooks still waiting for a second pair of eyes. Attesting a workbook is the review work this
PR is asking for. A workbook nobody has checked is a test that only proves the engine is
self-consistent with one person's spreadsheet.

### Spec sections
§3.3 and §3.8 of `roadmap/cost-spec-v2.md` (the worked-example contract and the attestation
fingerprints); the areas above map onto §4.4, §5 and §4.1–§4.2 of `cost_spec.md`.

## Push

```bash
git push origin cost/06-worked-examples

gh pr create --base cost/05-io-cli --head cost/06-worked-examples \
  --title "Cost module stack 6/8: hand-computed worked-example library" \
  --body-file roadmap/pr_descriptions/06-worked-examples.md
```


tests/worked_examples/discounting/eac_annuity_factor_20y_3pct.xlsx
valentinjanser, line 1: The naming is inconsistent with the one on Wikipedia, for instance.

Leaving out the negative sign:
EAC := NPV / "present value of annuity factor"
<=>
EAC = NPV * "loan repayment factor"

So, we might want to rename our factor.

valentinjanser:
General: Reviewing Excel sheets is terrible, roughly 30 seconds of pure git/file management per file.
If another way to do this exists, I would prefer that for next time. Maybe Google Sheets (one shared document with many tabs)?

tests/worked_examples/end_to_end/heat_pump_investment_and_bill.xlsx
valentinjanser, line 1: Our convention was to view "year 0" as undiscounted.
Therefore, fields such as "nominal_maintenance_year_1_in_euro" should be renamed, either to "undiscounted" (my preference) or to "..._year_0_..."

valentinjanser, line 1: The note "nominal, not discounted: the liquidity view of year 1" is misleading. The figure actually holds for all years, the nominal cashflow is constant for 1 <= year <= 10.

tests/worked_examples/end_to_end/heating_levy_559e_mixed_package.xlsx
valentinjanser, line 1: Not relevant in this (unrealistic) case, but remember that: 
- You have to subtract 15% of the investment from the levy calculation to account for the heating system having to be changed anyway (if it was already old).
- You can have a levy above 0.5€/m2 after 6 years.
- The rent still has to adhere to the limits as usual (not way above the typical rent for comparable objects), which makes the calculation less optimistic from a landlord's viewpoint if the rents are already on a high level.

tests/worked_examples/end_to_end/replacement_and_residual_value.xlsx
valentinjanser, line 1: Bad "programming" style: B22 uses a hard-coded 8 instead of the lifetime_in_years.
B23 as well, and here a modulo operation may be adequate (but watch out for horizons > 16 years...)

valentinjanser, line 1: Critical: I believe the residual value calculation to be erroneous. 

nominal_residual_value_year_12_in_euro := 4/8 * nominal_replacement[_cost]_year_8_in_euro
where 4 is the remaining lifetime and 8 is the total lifetime. 

Now, the spreadsheet does:
npv_residual_value_in_euro := nominal_residual_value_year_12_in_euro/(1+interest_rate)^horizon_in_years
So half the nominal value **from year 8** is discounted by 12 years.

I expected the nominal value from year 8 to be escalated by 4 years, i.e.:
nominal_residual_value_year_12_in_euro := -nominal_replacement_year_8_in_euro*(2*lifetime_in_years-horizon_in_years)/lifetime_in_years * (1+investment_price_escalation_rate)^(horizon_in_years-lifetime_in_years)

tests/worked_examples/financing/loan_10y_3pct.xlsx
valentinjanser, line 1: Contrary to the "loan_interest_only_bullet_..." file, there is no loan_type in the inputs here.

tests/worked_examples/financing/loan_partial_share_15y_5pct.xlsx
valentinjanser, line 1: Again, the loan_type is not explicitly specified.

tests/worked_examples/subsidies/combined_rate_cap_pro_rata.xlsx
valentinjanser, line 1: The description does not cover all the details, e.g., the measure cost.
The interest rate is specified, but unused.

tests/worked_examples/subsidies/grant_versus_tax_credit_choice.xlsx
valentinjanser, line 1: Actually, this file only computes variant B. Variant A is not to be found anywhere in the "EXPECTED" section. The description just says the grant yields 4000 € (which is correct), but this is not to be found anywhere in the structured data sections.

tests/worked_examples/subsidies/share_with_absolute_cap.xlsx
valentinjanser, line 1: According to the calculations, the description should be "30,000 EUR per dwelling unit" (but there is only one dwelling).

tests/worked_examples/tariffs/dynamic_spot_integral.xlsx
valentinjanser, line 1: Variable naming seems arbitrary: You would have to compare energy_working_in_euro with volume_effect_in_euro to see whether a dynamic/traditional contract is more fitting.
Also, there is no further markup on the spot price? Or is this included in the integral of power * "spot price"?

mean_energy_price_in_euro_per_kwh seems not to include the standing charge. Why not also call it something with "working"?

tests/worked_examples/tariffs/flat_household_bill.xlsx
valentinjanser, line 1: As above: mean_energy_price_in_euro_per_kwh should maybe contain "working".

valentinjanser (commented):
Review results: The possible calculation error in end_to_end/replacement_and_residual_value.xlsx should be cleared up before continuing.

tests/worked_examples/end_to_end/heat_pump_investment_and_bill.xlsx
nicolaigoelz, line 1: Hardcoded "12" in cell B35. Use instead months_per_year variable as it is done in end_to_end/heating_levy_559e_mixed_package.xlsx

tests/worked_examples/financing/loan_10y_3pct.xlsx
nicolaigoelz, line 1: Mismatch of filename (loan_10y_3pct) and name (loan_10y_3pct_round_numbers)

tests/worked_examples/end_to_end/replacement_and_residual_value.xlsx
valentinjanser: This seems to match the definition from VDI 2067, page 17, formula (3). Therefore, the residual value is the discounted book value (according to German business book keeping law) and not necessarily the actual value. 

Why do I say the latter? Imagine hyperinflation, r=q=100%.

In year 0, you buy the machine for 1000 €.
After 8 years, it will have to be replaced for 256,000 € nominal, 1000 € npv. Logical.
However, after 4 years, it will be worth 500 € NOMINAL, 31.25 € npv.

Looking at the NPV, the machine lost 96.875% of its value in the first half of its life.
I am unsure whether this way of calculation represents what we want to model.

nicolaigoelz:
Number formatting is applied inconsistently across the files, in places counterintuitively (display only, value is not affected). Can be fixed to avoid confusion 

Example:
eac_annuity_factor_20y_3pct.xlsx
Cell B17 (annuity_factor_via_pmt) is formatted as US dollars, but the annuity factor 
All other cells are `General` format

Suggestion:
set all cells to `General` to avoid confusion, or check the correct unit per cell and format accordingly

tests/worked_examples/end_to_end/heating_levy_559e_mixed_package.xlsx
nicolaigoelz, line 1: Please keep the sign convention: sign convention: cost positive, revenue / subsidy / residual value negative
subsidy_hestpump_in_euro and subsidy_wallinsulation_in_euro are positive values in this file.

tests/worked_examples/subsidies/share_with_absolute_cap.xlsx
nicolaigoelz, line 1: Support/award amount is the mirror image of a cost from the house-owner's perspective. 
Under the ledger sign convention (cost positive, revenue/subsidy negative) it would be negative.

tests/worked_examples/subsidies/per_dwelling_unit_tiered_cap.xlsx
nicolaigoelz, line 1: Support/award amount is the mirror image of a cost from the house-owner's perspective.
Under the ledger sign convention (cost positive, revenue/subsidy negative) it would be negative.

nicolaigoelz:
For all subsidies/* files check sign convention, cost is positive, award / subsidy negative

tests/worked_examples/end_to_end/pv_feed_in_over_10_years.xlsx
nicolaigoelz, line 1: "nominal_maintenance_year_1" should be renamed, as mentioned before
same for nominal_feed_in_revenue_year_1

nicolaigoelz:
Let's add a worked example with escalation ≠ 0 — currently all 23 examples use 0%, so escalation functionality is untested.

At 0% escalation, nominal vs. real doesn't matter. At ≠ 0% it does, so this example needs explicit, consistent nominal/real labeling. 
Per cost_specs, all rates are nominal, so let's use nominal rates and label them precisely and consistently across all examples.

nicolaigoelz:
Let's add a tariffs/ example for the CO2 price component of the retail bill, since none of the current examples cover it

noah80:
Thanks for the thorough round — every comment was checked against the workbooks, the runner (`tests/worked_example_runners.py`), the engine and the spec before acting. Summary of what changed and, in three places, why not.

## Fixed as reported

- **Hard-coded literals** — `replacement_and_residual_value` B22/B23 now name `lifetime_in_years` (`^lifetime_in_years`, `2*lifetime_in_years-horizon_in_years`), with the single-replacement bound (lifetime ≤ horizon < 2·lifetime) stated in the note; `heat_pump_investment_and_bill` gets a `months_per_year` input like the levy example.
- **Number formats** — three cells (not two) carried a leaked `[$$-409]` dollar format from `PMT`: `eac_annuity_factor` B17, `loan_10y_3pct` B17 and `loan_soft_30y_1pct` B17. Every value/tolerance cell in the library is now `General`, and the converter gained **rule 10** rejecting anything else, so it can't recur.
- **`loan_10y_3pct` name mismatch** — fixed, plus converter **rule 9**: metadata `name` must equal the file stem.
- **Missing `loan_type`** — both annuity loans now declare `loan_type: ANNUITY`; the runner errors explicitly on an unknown value.
- **Grant vs. tax credit, variant A missing** — added `scheme_a_award_in_euro = 0` (pinning that the grant *lost*), with the comparison quantified: 4,000 EUR upfront vs. 4,538.75 EUR discounted credit at 5 %.
- **Descriptions** — `combined_rate_cap_pro_rata` names the 20,000 EUR measure cost and explains that the interest rate is required by the solver objective but has no effect on two upfront grants; `share_with_absolute_cap` says "per dwelling unit".
- **Note wording** — "liquidity view of year 1" → "of any operating year (constant here — nothing escalates)".
- **Tariff labels** — `volume_effect_in_euro` → `bill_at_mean_price_in_euro`; `mean_energy_price_in_euro_per_kwh` → `mean_working_price_in_euro_per_kwh` (it excludes the standing charge) in all three tariff examples that used it.

## Sign convention — decided, and made honest

The solver returns awards as **positive magnitudes**; exactly one site (`subsidy_application.py`, `award.upfront_amount.as_revenue()`) turns them into negative ledger entries, and `UncertainValue.scale()` refuses negative factors precisely so no casual sign flip can swap a band's ends. The workbooks therefore keep magnitudes where the engine has magnitudes — but the boilerplate header that claimed the ledger convention on every file was overclaiming. Now: every `subsidies/*` workbook (and the levy example, which had no header) states that its quantities are award/deduction magnitudes, **and** each subsidy example additionally pins `support_ledger_npv_in_euro = −support_npv_in_euro` — the same support as it lands on the ledger, negative. The levy rows are renamed `subsidy_deduction_*` after §559 Abs. 2's *Abzug*. Where the ledger convention genuinely applies (feed-in revenue, residual value) the workbooks already honored it.

## Two findings we checked and are not changing

- **Residual value ("critical")** — the workbook is correct per spec §3.6 rule 3, `calculators/investment.py` and VDI 2067-1 formula (3): the last installation's price escalated to its *installation* year (never to the horizon), written down straight-line, discounted from year T. It is the standard's *book-value* convention, as the follow-up comment noted; a market-value residual would be a deliberate deviation from the method the whole engine claims. The workbook description and spec §3.6 now say this explicitly so it isn't re-derived next time.
- **`nominal_*_year_1` renames** — year 0 is the investment year (undiscounted); operating flows fall in years 1..T (`maintenance.py`: "entries for years 1..T"; the workbooks' own `present_value_factor_sum` note). `nominal_maintenance_year_1` therefore names the engine's actual year-1 entry, and stays exact under escalation, where "undiscounted" would become ambiguous across years. Both end-to-end workbooks now carry a note stating the convention.

Also: `annuity_factor` keeps its name (it matches `EconomicParameters.annuity_factor()` and VDI 2067's *Annuitätsfaktor*); the eac workbook notes it is the capital recovery factor, the reciprocal of the present-value annuity factor. The dynamic-spot markup question is answered by the workbook's own `markup_in_euro_per_kwh` input — the integral is pure spot, the markup is an additive component.

## New examples (both requested gaps confirmed)

- **`end_to_end/escalated_operating_costs`** — 3 % energy / 2 % general escalation over 10 years, the library's first growing nominal series. It pins the engine's exponent convention (recurring payments quoted at year-1 prices, `(1+g)^(t−1)`, so year 10 carries nine steps) with year-1 and year-10 nominals plus closed-form NPVs; energy escalation deliberately equals the discount rate so the energy NPV is exactly `T × year-1 / (1+i)`. (Investment escalation was already covered by the residual-value example.)
- **`end_to_end/co2_price_from_trajectory`** — the CO2 *price* component of the bill (figure 3 of `calculators/co2.py`, computed in `calculators/energy.py`, never part of `apply_tariff`): emission factor × kWh × exposure × the trajectory's price for the calendar year, step-interpolated (50/80/110 EUR/t at 2024/2027/2030), eight yearly nominals and an explicit eight-term NPV. Emission factor, exposure and trajectory come from the synthetic `XX` database and are copied into the workbook as inputs, so a change there fails the example rather than re-baselining it.

## Review workflow

The generated YAML next to each workbook *is* the readable rendering — inputs, values, tolerances and every formula as a `derivation` string. `roadmap/cost-spec-v2.md` §3.9 now documents the recipe: review the YAML diff; open the workbook only to spot-check or edit; never edit the YAML by hand.

## Levy law points

The 15 % deduction for a heating system that had to be replaced anyway **is** modelled (the §4.1 anyway-cost credit) and deliberately absent from this greenfield example, as its note says. The six-year §559 Abs. 3a window and the comparable-rent ceiling are genuine model simplifications and are now logged as such in `cost_spec.md` §6.4.

noah80:
Second round on this PR, answering the aggregated LLM review (41 deduplicated findings). Every one was checked against the code before acting; the verdicts and the owner's decisions are below.

## The review's best catch: the levy example could not fail

Both asserted levies in `heating_levy_559e_mixed_package` sit **exactly on their statutory caps**, so editing a rate in `allocation_DE_2024.json` left the suite green — while the runner's docstring promised the opposite. Confirmed by experiment, and closed:

- `ModernizationLevyOutcome` now carries the two resolved caps **and** the two uncapped `basis × rate` legs beside the capped ones. The workbook asserts the uncapped legs (2,100 and 3,200), which are what actually move when a rate changes.
- The runner reads the caps **off the engine outcome** instead of re-deriving `12 × living_area` itself — the cap formula is the engine's to own, so a change to it now fails the example rather than being reproduced by it.
- Proof: with `heating_levy_rate_per_year` edited 0.10 → 0.03 the example fails naming the leg (`Excel says 2100.0, the engine says 630.0`); reverted, it passes.

## Silent failures closed in the engine

- **Unknown time-of-use band names are refused.** Energy filed under a band the contract does not define was priced at the fallback band (the reviewer measured 1,225 EUR where correct banding gives 1,381). It now raises. Energy the meter assigned to *no* band — the total minus the banded sum — still bills at the fallback, because that is a modelled default and not a mismatch; the two cases are told apart by whether a key names an unknown band.
- **`ControllabilityDiscount.kind` is a `ControllabilityKind` enum**, joining the module's three sibling kinds. A typo used to leave both amount fields unread and the discount silently worth nothing; it now raises at construction. Round-trips by value, so no catalog file changes.
- **The solver publishes the objective it chose on** (`SubsidyDecision.discounted_support_in_euro`), so `support_npv_in_euro` is an engine figure instead of the test runner re-adding the awards under its own discounting. `effective_support_share` is now that discounted support over the measure cost — on the upfront total it reported **0 % support** for a scheme paid entirely as instalments, which is why the one example where they disagree never asserted it, and now does.

## Harness and tooling

- **The converter is a package.** At 1,147 lines it was more than twice the 500-line ceiling this PR's own runners module cites; `tools/worked_examples/` now has one module per seam (model, attestation, emitter, arithmetic, formulas, workbook, cli) with `tools/convert_worked_examples.py` kept as the entry point CI and the docs name. It still imports no `hisim`, because the drift check runs in a job that installs only `openpyxl` and `PyYAML`. All 25 fixtures regenerate **byte-identically** — verified by running the write mode, not just `--check`.
- **The two duplications are gone**: the §3.8 fingerprint protocol and the workbook-discovery predicate each have one owner in the package, imported by the tests instead of re-implemented under "must stay in sync" comments.
- **Tolerances are no longer rounded on the way into a fixture.** Rendering `abs_tol` at the value's decimal count changed what the example asserted: `0.025` was stored as `0.03` (looser than declared), `0.15` as `0.1` (stricter), anything below ~5e-13 as `0.0` — demanding bit-exact equality. Now emitted at full precision in plain decimal, since an exponent is a *string* to a YAML 1.1 loader.
- **The attestation switch is validated** against `{warn, error}`: `"ERROR"` or a typo used to leave the review gate silently disabled. `--check` now reports how many examples carry a valid attestation, `--list-unreviewed` gained the test it never had, and the enforcement path is exercised in both directions including the match branch no fixture reaches. The switch stays at `warn` — see below.
- **Mechanical, each its own silent failure**: an `OverflowError` from a workbook formula (`=x^5000`) no longer aborts the whole run with a traceback naming no file, and neither does any other unexpected per-workbook error; a NaN or infinite tolerance is refused before `math.log10`; `ExpectedRow` enforces `abs_tol > 0` itself rather than trusting its one caller; Excel's postfix `%` counts as arithmetic, so `=30%*40000` regains the stale-cache cross-check it silently skipped; two revenue label tokens matching nothing are gone; a fixture whose workbook was deleted is now caught, where it used to keep passing against the engine with its independent ground truth quietly gone; one malformed YAML fails as itself instead of taking down collection of the entire `-m base` suite; a non-positive measure cost raises instead of being absorbed into a zero support share; and every fraction-typed input is bounded to [0, 1], so `interest_rate: 3` for "3 %" can no longer pass the whole engine-vs-Excel chain by mis-scaling both sides equally.
- **The levy example moved to its own `modernization_levy` group**, so a group is decided by directory alone again rather than by which inputs an example happens to declare, and the group→entry-point map is 1:1.
- **`spec_section` names `cost_spec.md`** — documented in the converter and the spec rather than prefixed into 25 workbooks, since the generated header is compared byte for byte and editing it would rewrite every fixture.

## Not changed, with reasons

- **`loan_soft_30y_1pct` "redundant" with `loan_10y_3pct`** — mutation-coverage reasoning applied to a *reference* library. A 1 %/30-year soft loan is a distinct product a reader looks for, and a second hand-checkable point on the same branch costs nothing.
- **The double `openpyxl` load** is required: `data_only` is a load-time switch, so formula text and cached values cannot come from one parse.
- **The levy runner's second `build_timeline` and the discounting `filtered()` loop** are sub-millisecond; both fixes would mean engine plumbing for a test convenience.
- **`months_per_year` in the levy workbook** is a workbook-side derivation helper, exactly like `kilograms_per_ton` in the CO2 example — not a dead engine input.
- **Non-uniform runner signatures** (some take a `CostDatabase`): the dispatch lambda is adequate, and the group move already tidied that seam.
- **Enforcement stays `warn`.** The mechanism is sound; what was missing is the event it waits for — the human review round happening in this PR. `--check` now makes the attestation count visible, and the switch flips when the examples carry attestations. Worth noting against the "AI-drafted, therefore not independent" finding: reviewers on this PR independently recomputed several examples and their numbers matched, which is exactly the evidence that finding asks for.

noah80:
Second round on this PR, answering the aggregated LLM review (41 deduplicated findings). Every one was checked against the code before acting; the verdicts and the owner's decisions are below.

## The review's best catch: the levy example could not fail

Both asserted levies in `heating_levy_559e_mixed_package` sit **exactly on their statutory caps**, so editing a rate in `allocation_DE_2024.json` left the suite green — while the runner's docstring promised the opposite. Confirmed by experiment, and closed:

- `ModernizationLevyOutcome` now carries the two resolved caps **and** the two uncapped `basis × rate` legs beside the capped ones. The workbook asserts the uncapped legs (2,100 and 3,200), which are what actually move when a rate changes.
- The runner reads the caps **off the engine outcome** instead of re-deriving `12 × living_area` itself — the cap formula is the engine's to own, so a change to it now fails the example rather than being reproduced by it.
- Proof: with `heating_levy_rate_per_year` edited 0.10 → 0.03 the example fails naming the leg (`Excel says 2100.0, the engine says 630.0`); reverted, it passes.

## Silent failures closed in the engine

- **Unknown time-of-use band names are refused.** Energy filed under a band the contract does not define was priced at the fallback band (the reviewer measured 1,225 EUR where correct banding gives 1,381). It now raises. Energy the meter assigned to *no* band — the total minus the banded sum — still bills at the fallback, because that is a modelled default and not a mismatch; the two cases are told apart by whether a key names an unknown band.
- **`ControllabilityDiscount.kind` is a `ControllabilityKind` enum**, joining the module's three sibling kinds. A typo used to leave both amount fields unread and the discount silently worth nothing; it now raises at construction. Round-trips by value, so no catalog file changes.
- **The solver publishes the objective it chose on** (`SubsidyDecision.discounted_support_in_euro`), so `support_npv_in_euro` is an engine figure instead of the test runner re-adding the awards under its own discounting. `effective_support_share` is now that discounted support over the measure cost — on the upfront total it reported **0 % support** for a scheme paid entirely as instalments, which is why the one example where they disagree never asserted it, and now does.

## Harness and tooling

- **The converter is a package.** At 1,147 lines it was more than twice the 500-line ceiling this PR's own runners module cites; `tools/worked_examples/` now has one module per seam (model, attestation, emitter, arithmetic, formulas, workbook, cli) with `tools/convert_worked_examples.py` kept as the entry point CI and the docs name. It still imports no `hisim`, because the drift check runs in a job that installs only `openpyxl` and `PyYAML`. All 25 fixtures regenerate **byte-identically** — verified by running the write mode, not just `--check`.
- **The two duplications are gone**: the §3.8 fingerprint protocol and the workbook-discovery predicate each have one owner in the package, imported by the tests instead of re-implemented under "must stay in sync" comments.
- **Tolerances are no longer rounded on the way into a fixture.** Rendering `abs_tol` at the value's decimal count changed what the example asserted: `0.025` was stored as `0.03` (looser than declared), `0.15` as `0.1` (stricter), anything below ~5e-13 as `0.0` — demanding bit-exact equality. Now emitted at full precision in plain decimal, since an exponent is a *string* to a YAML 1.1 loader.
- **The attestation switch is validated** against `{warn, error}`: `"ERROR"` or a typo used to leave the review gate silently disabled. `--check` now reports how many examples carry a valid attestation, `--list-unreviewed` gained the test it never had, and the enforcement path is exercised in both directions including the match branch no fixture reaches. The switch stays at `warn` — see below.
- **Mechanical, each its own silent failure**: an `OverflowError` from a workbook formula (`=x^5000`) no longer aborts the whole run with a traceback naming no file, and neither does any other unexpected per-workbook error; a NaN or infinite tolerance is refused before `math.log10`; `ExpectedRow` enforces `abs_tol > 0` itself rather than trusting its one caller; Excel's postfix `%` counts as arithmetic, so `=30%*40000` regains the stale-cache cross-check it silently skipped; two revenue label tokens matching nothing are gone; a fixture whose workbook was deleted is now caught, where it used to keep passing against the engine with its independent ground truth quietly gone; one malformed YAML fails as itself instead of taking down collection of the entire `-m base` suite; a non-positive measure cost raises instead of being absorbed into a zero support share; and every fraction-typed input is bounded to [0, 1], so `interest_rate: 3` for "3 %" can no longer pass the whole engine-vs-Excel chain by mis-scaling both sides equally.
- **The levy example moved to its own `modernization_levy` group**, so a group is decided by directory alone again rather than by which inputs an example happens to declare, and the group→entry-point map is 1:1.
- **`spec_section` names `cost_spec.md`** — documented in the converter and the spec rather than prefixed into 25 workbooks, since the generated header is compared byte for byte and editing it would rewrite every fixture.

## Not changed, with reasons

- **`loan_soft_30y_1pct` "redundant" with `loan_10y_3pct`** — mutation-coverage reasoning applied to a *reference* library. A 1 %/30-year soft loan is a distinct product a reader looks for, and a second hand-checkable point on the same branch costs nothing.
- **The double `openpyxl` load** is required: `data_only` is a load-time switch, so formula text and cached values cannot come from one parse.
- **The levy runner's second `build_timeline` and the discounting `filtered()` loop** are sub-millisecond; both fixes would mean engine plumbing for a test convenience.
- **`months_per_year` in the levy workbook** is a workbook-side derivation helper, exactly like `kilograms_per_ton` in the CO2 example — not a dead engine input.
- **Non-uniform runner signatures** (some take a `CostDatabase`): the dispatch lambda is adequate, and the group move already tidied that seam.
- **Enforcement stays `warn`.** The mechanism is sound; what was missing is the event it waits for — the human review round happening in this PR. `--check` now makes the attestation count visible, and the switch flips when the examples carry attestations. Worth noting against the "AI-drafted, therefore not independent" finding: reviewers on this PR independently recomputed several examples and their numbers matched, which is exactly the evidence that finding asks for.

PR #566, opened 2026-08-18
String-valued config enums

noah80:
- **Make every config enum serialize as its member name**
  All 64 component config classes are dataclass_json-decorated, and
  dataclasses_json encodes an enum by its `.value` and decodes a typed enum
  field by calling `EnumClass(value)`. Giving each enum a string value equal
  to its member name therefore makes the on-disk representation
  self-describing with zero extra machinery: a scenario file now reads
  "DWD_TRY" instead of the opaque integer 1.
  
  Nine enums reached from config fields still carried integer values and are
  converted to `class Foo(str, Enum)` with `MEMBER = "MEMBER"`, `@unique`
  throughout:
  
    weather.WeatherDataSourceEnum
    heat_distribution_system.HeatDistributionSystemType
    heat_distribution_system.PositionHotWaterStorageInSystemSetup
    more_advanced_heat_pump_hplib.PositionHotWaterStorageInSystemSetup
    simple_water_storage.PositionHotWaterStorageInSystemSetup
    simple_water_storage.HotWaterStorageSizingEnum
    generic_pv_system.PVLibModuleAndInverterEnum
    generic_boiler.BoilerType
    generic_ev_charger.EVChargerMode
  
  Enums whose values were already meaningful strings (LoadTypes, Units,
  SimpleHeatSourceType, FluidMediaType, LpgDataAcquisitionMode,
  EnergyIntensityType) are untouched.
  
  This is a clean cutover with no backward-compatibility shims. Everything
  that existed only to keep the old integer encoding alive is deleted rather
  than adapted:
  
  - The `Union[SomeEnum, int]` annotations on config fields, factory
    signatures and build() parameters become the plain enum (11 sites in
    heat_distribution_system, 3 in simple_water_storage, 1 in
    more_advanced_heat_pump_hplib, 1 in generic_ev_charger).
  - `EVChargerControllerConfig.__post_init__`, which coerced a bare integer
    1-6 into an EVChargerMode member, is removed together with the now
    redundant `cast()` on the controller side.
  - The int-tolerant `config.heating_system in [FLOORHEATING, 2]` membership
    tests in `HeatDistribution.get_cost_capex` and the equivalent list test
    in SimpleHotWaterStorage become plain member comparisons.
  
  No value-dependent logic needed adapting: every use of these enums is a
  member comparison. In particular generic_pv_system never passes an
  ordinal to pvlib - `retrieve_sam` is called with hard-coded database
  names inside member branches - so the pvlib lookups are unaffected.
  
  Two tests are updated to the new contract:
  test_generic_ev_charger_config.py dropped its int-coercion cases and now
  pins the member-name values, the name-string lookup and a JSON round trip;
  test_heat_distribution_system.py replaces its integer sentinels.
  
  

- **Migrate scenario JSONs to the string enum values**
  The int enum values baked into the checked-in v1 scenario files no longer
  decode now that the enums are string-valued, so the data is rewritten
  rather than the decoder made int-tolerant. Old JSON is treated as broken
  data to be revised, not as an input format to keep supporting.
  
  The rewrite was scripted and is class-aware: each component entry names its
  config class in `config_full_classname`, so the script imports that class,
  resolves the field annotations (through Union/Optional) and rewrites only
  fields whose declared type is one of the converted enums, mapping the
  historic integer to the matching member name. Generic integer fields that
  merely share a name with an enum field are therefore left alone - notably
  `mode` on GenericHeatPumpControllerConfig and
  MoreAdvancedHeatPumpHPLibControllerSpaceHeatingConfig, which is a plain int
  and not an EVChargerMode.
  
  20 of the 23 files in system_setups/ changed, 118 field values in total:
  data_source (20), heating_system (26), position_hot_water_storage_in_system
  (29), module_database (17), inverter_database (17), boiler_type (8). Files
  are re-emitted with indent=4 and no trailing newline, matching
  hisim.json_generator, so the diff is confined to the changed values.
  
  The scenario JSONs under obsolete/ are deliberately not migrated: they
  reference component modules that were deleted long ago
  (generic_heat_pump_modular, generic_hot_water_storage_modular,
  generic_heat_source) and so cannot be loaded either before or after this
  change. Migrating them would only suggest otherwise.
  
  Verified by decoding all 239 component configurations across the 23
  scenario files through the same `from_dict` path the JSON executor uses,
  and by tests/test_hisim_main_json.py, which runs a real v1 scenario file
  end to end.
  
  Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
  

- **Add a round-trip proof for config enum serialization**
  New test module pinning the contract the string cutover establishes, so
  that a future IntEnum in a config class fails a test instead of silently
  producing an unreadable scenario file.
  
  Four properties are covered. Each converted enum carries its own name as
  its value and, since the cutover left no shim behind, rejects the legacy
  integer codes outright. Each config class owning an enum-typed field
  serializes those fields as readable strings and survives a full
  encode/decode round trip with the enum members intact. The converted
  configs are additionally driven through `json.dumps(config.to_dict())`,
  the path hisim.json_generator actually uses when writing scenario files,
  which only works because the members are str subclasses.
  
  A final sweep imports every module under hisim/components, walks all
  ConfigBase subclasses' dataclass fields (resolving Union/Optional
  annotations) and fails on any enum still carrying non-string values.
  
  Registry tables live inside a class rather than at module level, per the
  project's no-module-level-constants convention.
  
  One pre-existing limitation is documented in the registry rather than
  fixed here: SimpleHeatSourceType and FluidMediaType are plain (non-str)
  Enums with meaningful string values, so SimpleHeatSourceConfig cannot go
  through `json.dumps(to_dict())` at all. That predates this change and
  those enums were explicitly out of scope, so it is only recorded.


hisim/components/heat_distribution_system.py
valentinjanser, line 1: SERIE = "SERIE"
Maybe change to SERIAL?

hisim/components/more_advanced_heat_pump_hplib.py
```diff
@@ -71,9 +76,9 @@ class PositionHotWaterStorageInSystemSetup(IntEnum):
     No Hot Water Storage in system setup for space heating
     """
 
-    PARALLEL = 1
-    SERIE = 2
-    NO_STORAGE = 3
+    PARALLEL = "PARALLEL"
+    SERIE = "SERIE"
```
valentinjanser: Same here. "Series" would also be an option.

hisim/components/simple_water_storage.py
```diff
@@ -38,21 +38,34 @@
...
+    """Set Simple Hot Water Storage Position options.
+
+    Decides whether the storage sits parallel to or in series with the heat
+    source, which changes the component's input wiring. Every member carries its
+    own name as its value so that a serialized configuration spells the position
+    out instead of encoding it as an integer.
+    """
+
+    PARALLEL_TO_HEAT_SOURCE = "PARALLEL_TO_HEAT_SOURCE"
+    SERIE_TO_HEAT_SOURCE = "SERIE_TO_HEAT_SOURCE"
```
valentinjanser: Same here

k-rieck (commented):
- if the checks and pytests are ok, I agree with merging

valentinjanser:
1. The changes are very much welcomed. Makes the configurations much more readable and maintainable.
2. At some point, we will likely have to refactor our test base. The test of this PR, afaik, already overlaps with other tests (the round-trip testing of JSONs). However, I do not see this as a pressing problem now. Maybe create an issue?

PR #570, opened 2026-08-18
Structured component

noah80:
## Motivation

Component identity is currently three loose mechanisms: a `name` config field, a
`building_name` config field, and `get_component_name()`, which prefixes the two
**depending on the `multiple_buildings` simulation parameter** — the same config
produces different identities in different runs. Downstream, KPI/postprocessing
recovers structure by *parsing names back apart* (`component_name.split("_")[0]`,
substring matching against `DistrictNames` tokens).

This PR replaces all of it with a structured, first-class identity, as the
foundation for upcoming work: multi-family buildings (per-unit devices) and
district calculations (per-building KPIs) become data instead of naming
conventions, and the JSON scenario architecture v2 can reference components by a
well-defined key.

## What changed

- **`ComponentID`** (frozen, `dataclass_json`): `name` (required), `building`,
  `unit` (optional). The runtime identifier is the derived `key` — the present
  fields joined with `_` in the order building, unit, name. The key is
  derived-only and never parsed anywhere; ambiguous-tuple collisions are caught
  by the existing duplicate-name check at output registration.
- **All 85 config classes** carry `component_id: ComponentID` instead of the
  `name`/`building_name` pair; 115 default/scaled factories take
  `component_id: Optional[ComponentID] = None`; ~220 construction sites updated.
- **`get_component_name()` is now a pure function of the component**
  (`component_id.key`); `multiple_buildings` no longer influences naming.
- **KPI/postprocessing** grouping reads the structured fields: the
  `DistrictNames` substring matching (~45 sites) became exact membership tests
  via `DistrictNames.is_district(building)`, and `split("_")[0]`-style grouping
  became field comparisons.
- **The 23 checked-in scenario JSONs** were migrated by script (239 config
  blocks): `"building_name"`/`"name"` → nested `"component_id"`. No
  compatibility shim — old-shaped JSON fails to load.

## Result-neutrality (the correctness argument)

The derived key reproduces today's names exactly: bare `name` when no building
is set, `BUI1_Weather` when one is. An audit found `multiple_buildings=True`
**nowhere in live code** (one unit test only) — every `building_name="BUI1"` in
the repo was decorative and suppressed by the flag. All configs therefore
migrate to `building=None`, and every runtime name, result column, and grouping
key is unchanged.

Verified: 980 base tests green (incl. `test_hisim_main_python`), the
extendedbase meter/EMS/PV set green, `test_hisim_main_json` green (migrated
JSONs through the real executor), new `tests/test_component_id.py` (23 tests,
incl. district-style prefix cases), flake8 zero new violations, mypy clean over
all 128 files.

## Reviewer notes

- **`ComponentID.building_label`** is the one piece of new semantics:
  postprocessing keys per-building groups by the literal `"BUI1"` (hardcoded for
  the webtool JSON at `postprocessing_main.py`), so `building_label` returns the
  building when set and the default label `"BUI1"` otherwise, keeping group keys
  byte-identical. Worth a close look.
- `DistrictNames.is_district` is an exact-membership *tightening* of the old
  substring test (documented in its docstring); the two agree on every value in
  live and obsolete setups.
- Out of scope on purpose: `ArcheTypeConfig.building_name` (building-sizer
  interface contract, not component identity); `obsolete/` scenario JSONs
  (reference deleted modules).


hisim/component.py
```diff
@@ -82,8 +179,15 @@ def __init__(
         sankey_flow_direction: Optional[bool] = None,
         output_description: Optional[str] = None,
         source_component_class: Optional[str] = None,
+        component_id: Optional[ComponentID] = None,
```
valentinjanser: Is there a specific reason why this has to be Optional?

valentinjanser (approved):
In general, I think this change of unifying multiple naming schemes is beneficial to the usability of HiSim, especially looking at the scenario descriptions.

However: We should very soon work on the code style, multiple tests failing...

k-rieck (commented):
approved but please fix tests and code quality before merging

PR #577, opened 2026-08-21
Building cleanup phase 1: the characterization harness (plus the phased spec and findings log)

noah80:
Phase 1 of the building cleanup (plan of record: `roadmap/building_cleanup_spec.md`, in this diff): a three-layer characterization harness for `building.py`, as a pure addition — nothing under `hisim/` changes. The harness is the referee for every later phase: the mechanical package split, the per-class refactors, and the eventual decision on the 2,000-line component itself.

**Layer 1 — `BuildingInformation` snapshot.** Every one of the 2,974 TABULA building codes is instantiated and all 62 derived attributes compared exactly (floats bit-for-bit) against a committed golden. 124 codes that currently crash (`ZeroDivisionError` in the window-area scaling) are pinned as recorded behavior, not failures. Five codes additionally run under config-override variants so every override branch is covered. Runs in ~44 s.

**Layer 2 — one-day `Building` snapshot.** 96 timesteps of synthetic deterministic inputs straight through the `i_simulate` loop (no weather files, no simulator), all 16 output vectors pinned exactly, one save/restore round mid-run, plus a scaled-config variant. Runs in ~1 s.

**Layer 3** is the existing golden/parity CI, declared as the slow backstop.

Regeneration only via `HISIM_REGENERATE_BUILDING_GOLDENS=1`, and always from the complete sweep — a `-k`-narrowed run cannot write a partial golden. The harness was verified negatively: deliberately perturbing constants in `building.py` turns it red with precise diffs (then restored byte-identically).

**Also in this PR for review:** `roadmap/building_cleanup_spec.md` (the full phased plan — please review the phase definitions, especially the phase-6 draft and its go/no-go framing) and `roadmap/building_cleanup_findings.md` (everything the harness and the follow-up phases have pinned so far — including the 124 crashing TABULA codes, the `sum()`-vs-chained-`+` float discovery, and a 13-month-old latent tuple-unpack bug that was never noticed because its results were never read). Findings entries referencing phases 2–4 describe work that follows as stacked PRs on top of this one.

Gates: base suite 3,988 passed (988 baseline + the 3,000 new tests), buildingtest 11 passed / 1 xfailed, prospector 0 messages, mypy clean.


k-rieck (approved):
- how about adding an error tolerance in the new golden building pytests as for the original golden refs?

PR #578, opened 2026-08-21
Building cleanup phase 2: the mechanical package split (pure moves)

noah80:
Phase 2 of the building cleanup (spec: `roadmap/building_cleanup_spec.md`, shipped with phase 1). **Stacked on #577 — review only the top three commits.**

The 3,082-line `building.py` becomes the `hisim/components/building/` package, as **pure moves** — class bodies verified byte-identical, `git mv`-based so blame survives:

| file | lines | contents |
|---|---|---|
| `__init__.py` | 75 | re-exports all five public names; the public import path `hisim.components.building.*` stays stable |
| `config.py` | 156 | `BuildingConfig` |
| `building.py` | 2,071 | `Building` + `BuildingState` |
| `window.py` | 150 | `Window` |
| `information.py` | 713 | `BuildingInformation` |

Split by responsibility, not one-class-per-file: `BuildingState` (~30 lines, used only by the component's save/restore) stays with `Building`.

**What reviewers should check:**
- The only non-move additions are module docstrings and per-file imports, plus one sanctioned cycle resolution: `BuildingConfig.get_main_classname()` needs `Building`, so that single reference became a method-local import.
- **The module path is wire format**: `get_full_classname()` derives from `__module__`, so the second commit regenerates the committed scenario JSONs — 19 of 23 files change by exactly two classname strings each, zero value changes. Old JSON strings (`hisim.components.building.Building`) keep resolving through the package re-exports, verified both directions. The `wetterdienst`-dependent setup could not regenerate locally (three dependency failures deep); its two strings were substituted by hand and the freshness CI is the final verifier.
- The third commit fixes the one thing the regeneration missed: a test pinning the old classname string.
- No scanner needed changes: `check_config_attrs`, the v1 executor's class resolution and the code-overview generator all handle the package layout.

**The proof:** the phase-1 harness (3,000 characterization tests + the one-day snapshots) passes with **untouched goldens** — that is the phase's entire claim. Full base suite: 3,988 passed.


k-rieck (commented):
- CI checks are missing
- but otherwise approved

valentinjanser (approved):

PR #582, opened 2026-08-21
Move ConfigBase, ComponentID and DisplayConfig into a hisim/config pa…

noah80:
…ckage

The three classes every component configuration is built from lived in hisim/component.py, the component *runtime* module. That placement is what forces every future config-layer addition (named presets, sizing laws) either into the already 766-line component module or into a module that closes an import cycle with it. This commit creates the layer they belong to and moves them there, with no behavior change of any kind.

- hisim/config/base.py holds ComponentID, ConfigBase and DisplayConfig with verbatim bodies (checked line-by-line against the removed block). The module imports nothing from the rest of HiSim, which is the property that makes the package a genuine bottom layer.
- hisim/config/__init__.py exports the three names and states the layering rule the package exists to enforce.
- No compatibility alias is left behind in hisim/component.py. Instead all 169 call sites now import from hisim.config directly, so the layering is visible where it matters and cannot silently rot back into the old path. component.py itself uses the classes through a `cfg` module alias, the repo's usual spelling for a cross-module dependency, so `hisim.component.ConfigBase` no longer resolves at all.
- docs gains a config page (the three classes would otherwise have dropped out of the generated API reference) and CLAUDE.md gains the layer.

Purely mechanical: 499 insertions against 633 deletions are import lines and `cp.` prefix removals for exactly these three names. Nothing about the wire format changes -- scenario JSONs name concrete config classes under hisim.components.*, never these base classes (verified over two of them).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -54,6 +54,14 @@ UTSP_API_KEY
 ### Core simulation loop (`hisim/simulator.py`)
 `Simulator` owns a list of `ComponentWrapper` objects and runs `run_all_timesteps()`. Each time step iterates through all wrapped components, calling `i_simulate()` on each, until all outputs converge (checked via `SingleTimeStepValues.is_close_enough_to_previous()`). After convergence, `PostProcessor` is invoked.
 
+### Config layer (`hisim/config/`)
+The bottom layer, imported by everything else and importing nothing from the rest of HiSim.
+`hisim/config/base.py` holds `ComponentID` (the structured component identity),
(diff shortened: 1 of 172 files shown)


k-rieck (commented):
- approve
- but I think DisplayConfig is a leftover of Kevin's former webtool approach and we might want to remove this. The "pretty name" we need for plots though, but that is already implemented in the chart making modules in postprocessing/ as far as I know

k-rieck (approved):

PR #584, opened 2026-08-23
Add the cache service specification

noah80:
Spec for a shared cache of precomputed HiSim artifacts (weather, LPG profiles, PV yields, solar gains) served by the RenoVisor REST server: static producer functions with single-DTO signatures, fully automatic import-closure fingerprinting for cache keys, content-addressed curated datasets, tiered lookup (local dir -> server -> shared dir -> compute), and OIDC/token write authentication.

### DIFF
--- a/roadmap/cache_service_spec.md
+++ b/roadmap/cache_service_spec.md
@@ -1,0 +1,466 @@
+# HiSim Cache Service — Specification (v1 draft)
+
+A shared cache service for precomputed HiSim artifacts (LPG occupancy profiles, weather
+preprocessing, PV yields, heat-pump maps, solar gains, …), served over REST by the existing
+RenoVisor C# server, plus a client library inside HiSim that every component uses instead of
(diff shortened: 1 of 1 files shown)


nicolaigoelz:
Looks very promising. Couple of open points for discussion:
- Non-public datasets: E.g., Austrian weather data, or (possiby) AustrianStandards Norm-Input-tables are non-public. Specification states in §6 that reads are anonymous on both ends. Thus, we need to set up a "internal-only" area
- 409 conflict comparison (already existing, different content): Maybe it makes sense to include a numeric tolerance comparison? And is the rejected blob retained for (manual?) comparison?
- DOI & citability not possible with our current approach, would be an advantage of archive server like zenodo. Something we could still use?
- Prposal: Keep DWD-Aachen as only "real dataset" in repo; add minimal synthetic fixtures to test other parser
- Weather reader cleanup: some differences / complications in current weather implementation, that should be cleaned up at one point

PR #586, opened 2026-08-23
P1 — Sizing kernel: presets, AUTO fields and the fact binding rule

noah80:
**Title:** P1 — Sizing kernel: presets, AUTO fields and the fact binding rule

Stacked on #589 (roadmap reorganisation and the declarative-energy-systems requirements); retargets to `main` when #589 merges.

Implements phase P1 of the declarative energy systems epic:
`roadmap/declarative_energy_systems/epic.md` · requirements `p1_sizing_kernel_requirements.md` · design `p1_implementation_spec.md` (incl. the PR-A/PR-B amendments) · plan `plan.md` §P1.

## What changed since the last review

The earlier revision of this PR bound cross-component sizing facts through a three-way scope (global pool, connection-adjacent lookup, flat-pool fallback, file-level pre-seed). Review found that too implicit once two components of one class share a house. That mechanism is gone. In its place, one rule (requirements R4.3, epic E2):

> Every sizing fact a component declares is addressable as `<name>.<fact>`. A law's bare fact binds only if exactly one enabled config declares it; otherwise the consumer names its source (`resolve_all(configs, sources={consumer: {fact: "<name>.<fact>"}})`) and the engine errors listing the candidates together with the mapping to paste. Providership is decided from class declarations before anything is computed — a provider whose value is `None` still counts — so the outcome cannot depend on resolution order.

## Contents (21 commits)

- **Kernel** (`hisim/config/`): `laws.py` (law algebra; new `Self("field")` sibling term with intra-config ordering, `Many(...)` declared but raising until a real consumer exists, `Cardinality`), `context.py` (`SizingContext`, `Size` terms), `sizing.py` (`AUTO`, `sized_field(note=)`, `resolve_config` with dependency-ordered fields and the `OwnFields` view), `engine.py` (registration → validation → fixed point; the binding table and error catalogue E-01…E-08), `report.py` (`ResolutionReport`: sweeps, lookups with candidates, unconsumed facts), `presets.py` (`@preset`/`@constructor` decorators on classmethods; `GenericBoilerConfig.preset_condensing_gas("Boiler")` is typed as the config class and a misspelled preset is a mypy `attr-defined` error, no plugin; mandatory `preset_`/`for_` method prefixes), `introspection.py` (`describe_config`: fields, presets with pinned/AUTO, sizable fields with law, facts and cardinality, constant-law marker, facts provided).
- **Central check**: `Component.__init__` rejects configs still carrying `AUTO`, naming field and law.
- **Presets carry no instance name**: presets and constructors take the name as first argument (`GenericBoilerConfig.preset_condensing_gas("Boiler")`, `BuildingConfig.for_tabula_code("Building", building_code=…)`); 110 call sites converted with component names preserved, so no result columns change.
- **Pilots** converted, legacy factories deleted: `GenericBoilerConfig` (7 presets; pellet/wood-chip minimum = `Self("maximal_thermal_power_in_watt") * 1/12`, so a pinned maximum keeps the minimum consistent), `HeatDistributionConfig`, `EMSConfig`, `BuildingConfig` (one preset, `standard`, delegating to the first constructor `for_tabula_code`).
- **Tests**: `test_sizing.py`, `test_sizing_engine.py` (rewritten: binding table, ambiguity, permutation independence, sibling reads, cycles, provenance), `test_presets.py` (incl. an in-test mypy run for a misspelled preset and a wrong-typed constructor argument), `test_config_contracts.py` (over every component config: fact-name collisions, one-vs-many, `Self` names and cycles, preset builds, naming convention on wire names, wire-format pin incl. `for_tabula_code`).
- **Docs**: the superseded plan and review agenda deleted (their items are answered in the epic's decision register and P1 §11); `system_docs/config_defaults_spec.md` §8.4/§8b marked superseded.

## What a reviewer must decide

1. The binding rule as `resolve_all(configs, sources=…)` with providership from declarations over the given set; seed context counts as a provider named `<seed>` (impl spec §3.2).
2. Sibling reads via `Self("field")`, final value pinned or sized, validated at first resolve and by the contract test (§3.3, DQ1 amended).
3. Presets and constructors as decorated classmethods with mandatory `preset_`/`for_` prefixes, the instance name as first argument (§3.4 as amended 2026-08-26).
4. `Many(...)` as a declared-but-raising hook until the first real consumer (epic EQ2).
5. Preset names are unchanged except `german_single_family_home → standard`; the naming convention is decided (five rules + A1–A3, enforced by the contract test); all P1 questions are decided.

## Gates

`pytest -q tests/test_sizing.py tests/test_sizing_engine.py tests/test_config_contracts.py tests/test_presets.py` → 79 passed · `pytest -m base -k "boiler or heat_distribution or energy_management or building or json"` → 3179 passed, 1 skipped · `pytest -m jsonconfig` → 22 passed, 2 skipped · full `pytest -m base` → 4118 passed, 1 skipped (5:40) · mypy (`hisim/config` + 4 pilots) → no issues · flake8 (120) → clean · `grep -rn "FactScope\|preseeded_facts\|adjacency" hisim/` → empty · pilot resolved values byte-identical before/after the kernel rework · zero `*.scenario.json` fixture changes.

## Review guide (from `p1_implementation_spec.md` §12)

Look first at `engine.py::_bind` (the binding table, row by row), `sizing.py::resolve_config` (dependency ordering and the `OwnFields` view), `presets.py` (decorator signature preservation, prefix enforcement; no literal `ComponentID(name="…")` inside a preset body). Reject: any fallback read on a fact pool; any `if len(providers) > 1: pick …`; a preset carrying a literal name; any `from hisim.components` under `hisim/config/`. Out of scope here: file syntax and `sizing_sources` parsing (P2), many-cardinality evaluation (parking lot), converting classes beyond the pilots (P4), the `describe` CLI (P2).


hisim/config/laws.py
```diff
@@ -0,0 +1,294 @@
...
+    """Turns whatever ``sized_field(rule=...)`` or :func:`law` received into a law object.
+
+    Accepts a ready law, a callable (which **must** declare its reads), or any
+    other value (wrapped as a constant law) — the three spellings the spec allows. The
+    ``reads`` entries may be ``Size.*`` terms or plain fact names.
+    """
+    if isinstance(rule, SizingLaw):
+        return rule
+    if callable(rule):
+        if reads is None:
```
valentinjanser, line 474: I would find it nicer if the naming was more uniform, i.e., name it facts_read instead of "reads" in accordance with the function in the SizingLaw class.

hisim/config/contributions.py
```diff
@@ -0,0 +1,77 @@
...
+    #: contributions, e.g. ``BuildingConfig.SIZING_CONTRIBUTIONS = (FactContribution(...),)``.
+    #: The engine reads the declarations through this name.
+    CLASS_ATTRIBUTE: ClassVar[str] = "SIZING_CONTRIBUTIONS"
+
+    facts: Tuple[str, ...]
+    compute: Callable[[Any, SizingContext], Mapping[str, Any]]
+    reads: Tuple[str, ...] = ()
+    scope: str = FactScope.GLOBAL
+
+    def __post_init__(self) -> None:
```
valentinjanser, line 54: It maybe makes sense to make the default for “reads” (may be renamed to prerequisites) to None and then check in the post_init whether the reads has been explicitly replaced. With the current state, I expect the “reads” to be forgotten to be set, leading to errors later on.

I would also like to see more documentation on what exactly the "reads" on the contributions does and how it's different from the "reads" on the laws.

Other than that, the changes look promising.

valentinjanser (approved):

roadmap/design_review_agenda.md
```diff
@@ -0,0 +1,297 @@
...
+
+- **(a1) Restricted own-field reads — plain (non-sizable) siblings only.**
+  `sized_field(..., reads_own=("share_of_maximum_pv_potential",))`; the engine feeds
+  the sibling values to the law. No cycles by construction (a plain field is never
+  law-computed, so nothing that reads it can loop back), no intra-config ordering
+  (plain fields hold their final values before resolution starts), and naming a
+  *sizable* field in `reads_own` is rejected at import. Fully covers the PV case —
+  the trap disappears, because an overridden share is simply what the law reads.
+  Does **not** cover B7.
+- **(a2) Full own-field reads via the existing fixed point, at field granularity.**
```
valentinjanser: With the given examples, I could think of certain workarounds (e.g., change from minimum_power to share_of_minimum_power_rel_to_max); however, to not have to rely on planned workarounds in the future, I see (a2) as the most sensible approach. 
(a1) again introduces an asymmetry (and therefore needed distinction) between sized and non-sizable fields, which I don't like.
(b) is not a solution at all, in my opinion.

tldr: a2

roadmap/design_review_agenda.md
```diff
@@ -0,0 +1,297 @@
...
+
+`Catalog.__getattr__` returns `Any`, so every helper returning a preset needs a manual
+annotation and preset-name typos get no static check. The sweep would replicate that
+~70×. One obstacle found during the rebuild (`roadmap/random_findings.md`): the natural
+spelling `presets: ClassVar[Catalog["GenericBoilerConfig"]]` is rejected — a `ClassVar`
+may not contain a type variable — so plain generics don't just drop in. Candidate
+mechanisms for phase 3: a `__class_getitem__`-based `Catalog[SomeConfig]` declared as a
+plain class attribute (not `ClassVar`), or a descriptor/protocol spelling. To decide
+*here* is only: **must preset access be statically typed before the sweep** (and typos
+caught by mypy), with the mechanism left to phase 3?
```
valentinjanser: I vote for *yes*

roadmap/design_review_agenda.md
```diff
@@ -0,0 +1,297 @@
...
+
+### B9. Which factories become presets at all — the Weather lesson
+
+Not every default factory should become a preset. `WeatherConfig`'s factory is a
+*lookup* over several dozen `LocationEnum` members — an open identifier space, not a
+variant set; converting it exhaustively would mint dozens of meaningless presets. The
+spike's pattern: presets only for the cases the repo actually uses (`aachen`,
+`seville`), plus a plain constructor for the general case (`for_location(...)`, named
+so no default-discovery heuristic picks it up). Same logic applies to TABULA building
+codes and LPG household definitions.
```
valentinjanser: I do not like this. 
Yes, *currently*, the repo only uses Aachen and Seville, maybe, but this has to be expected to change.
I would not make an exception for two “random” cities.
Do the for_location(...) pattern; it should be usable enough to not pose a problem for the cases where we use Aachen/Seville in this repo.

roadmap/design_review_agenda.md
```diff
@@ -0,0 +1,297 @@
...
+such notes should reach the realized record and audit artifact, they must be data:
+e.g. `sized_field(..., note="VDI 4645")` or a per-preset note.
+
+**The question:** do we want the convention at all, and in which shape (per-field,
+per-preset, both)? It is small machinery but a repo-wide authoring convention — the
+sweep would apply it ~70 times, and half-maintained citations are worse than none
+(a wrong "VDI 4645" misleads with authority).
+
+**Recommendation on file:** decide the shape now, make it strictly optional, populate
+only where a source is actually known — never invent citations to fill the field.
```
valentinjanser: I wonder whether it would really be a big problem to leave this out for now, and implement it later. (This was my first instinct)

If it has to be implemented, I agree with the recommendation and want to note that for some fields, multiple notes will have to be added. Therefore, instead of
`sized_field(..., note="VDI 4645")`
maybe we should do
`sized_field(..., notes=["VDI 4645"])`
?

roadmap/design_review_agenda.md
```diff
@@ -0,0 +1,297 @@
...
+contribution instead of resolving per loop — loud, not wrong, and defensible while a
+building has exactly one loop; but two boilers *are* already supported by the
+CONNECTED mechanism, so the asymmetry is a modeling decision, currently implicit.
+
+**To decide:** the normative scope rule the sweep applies uniformly — proposal:
+*a fact describing the scope itself (building physics, the single heat-distribution
+loop) is GLOBAL; a fact describing one device of possibly several (a generator's power
+band) is CONNECTED* — and whether the HDS controller's facts stay GLOBAL under that
+rule (i.e. one loop per building is a modeling commitment) or flip to CONNECTED now,
+before the sweep multiplies the pattern.
```
valentinjanser: For most components, you could argue that there *could* be two in one simulated scenario.
I wonder whether we need the global scope AT ALL. In the end, the complex dependency tree is, either way, created and traversed.

roadmap/design_review_agenda.md
```diff
@@ -0,0 +1,297 @@
...
+### C10. The UTSP `JsonReference` key-spelling bug
+
+The hand-authored v2 scenario fixtures spell the LPG connector's nested references in
+snake_case (`"household": {"name": …, "guid": {"str_val": …}}`), but the utspclient
+dataclass fields are `Name`/`Guid`/`StrVal`. dataclasses_json matches nothing, so they
+silently deserialize to empty references — invisible only because the fixtures run on
+the predefined profile; under `USE_LOCAL_LPG` household selection would be broken.
+
+**To decide:** fix the key spelling in the fixtures and verify one local-LPG run —
+as its own commit with before/after evidence (it changes what the files build).
```
valentinjanser: If there is ANY way to preserve the snake case, I would prefer that.
Also, I would like a real check-up on this, because as I initially implemented the JSON executor (and adapted some dataclass decorators), everything worked with snake_case. My workaround was then flagged by the LLM to be "bad style" and I wonder whether the LLM then actually went on and broke compatibility 😆 

tldr: check whether this is actually the case, then find a solution without changing the scenario definition schema

roadmap/design_review_agenda.md
```diff
@@ -0,0 +1,297 @@
...
+### C11. Buffer storage sized from building load vs generator power
+
+The legacy gas/oil/pellet/wood-chip setups size the buffer storage from the *building
+heating load* although the parameter is named after the *generator power* — a different
+number once the boiler is sized (≈1.1 × max(load, DHW)). The presets preserve this
+byte-identically; the heat pump chain genuinely reads the generator's contributed power
+(the two coincide there, because the heat pump's law is a pass-through of the load).
+
+**To decide:** bless the status quo explicitly, or schedule combustion buffers sized to
+actual boiler power as a deliberate physics change with result diffs.
```
valentinjanser: The result may differ; I would not mind.
In fact, the buffer size seems to largely depend on the heating system's rated power:
https://www.jnodenergy.com/news/choosing-the-right-buffer-tank-size-for-your-heat-pump/

However, e.g., in the case of the heat pump, the buffer size should be increased if floor heating is used to prevent frequent cycling. It would be advantageous to also add this dependency in one step and verify the credibility of the new results.

roadmap/design_review_agenda.md
```diff
@@ -0,0 +1,297 @@
...
+### C12. `heating_system` sizes to a constant, not to the building
+
+`HeatDistributionControllerConfig.heating_system`'s law is the constant `FLOORHEATING`:
+`"AUTO"` there means "the usual choice", not "derived from the building" — TABULA
+carries no heat-distribution-type information, so building-driven inference would be
+invented physics.
+
+**To decide:** bless that semantics of AUTO on this field (plus one documenting
+sentence), or commission a building-age→radiator heuristic — a result-changing decision
+for old buildings.
```
valentinjanser: Later on, literature research on correlations between the existence of floor heating and certain building parameters could really be interesting, but I don't think this has to be a topic now.

roadmap/design_review_agenda.md
```diff
@@ -0,0 +1,297 @@
...
+system setup. Three are zombies whose default connections import modules deleted long
+ago (`controller_l1_building_heating`, `controller_l1_heatpump`,
+`controller_l1_generic_runtime`); six are defective (`generic_battery` ×2,
+`generic_ev_charger` ×4: missing mandatory output descriptions, or defaults naming
+nonexistent database entries); five are legitimately data-dependent and keep their
+skips.
+
+**To decide:** delete outright, or move to `obsolete/` (the `SmartController`
+precedent). Executing this in phase 3 shrinks the sweep — nobody converts a class that
+is about to be deleted.
```
valentinjanser: @k-rieck What do you think? Can we move them all to obsolete/?

roadmap/design_review_agenda.md
```diff
@@ -0,0 +1,297 @@
...
+
+**Decision:** _open_
+
+### Q5. The preset naming convention (names are wire format)
```
valentinjanser: 1. I agree
2. Not really elegant, but right now, I cannot think of a particular example where this would clash (i.e., there has to be a number in the bare variant name, or there is no number in the size rating suffix). Also, I assume "concrete catalog device" does not refer to an actually existing model from our database (with manufacturer, model, etc.) but rather some specific fictional device?
3. Then we should fix whether this first preset is always nominal or sizable
4. Yes
5. Ok

k-rieck (commented):
Comments on decisions:
- B6: I prefer A1. A2 seems more complex and I don't see yet why we would need that. In general, before adding a higher abstraction or complexity layer, it would make sense to identify the relevant use cases first, and then decide whether that complexity is even necessary and worth it. The example and argument used for A2 is the minimum thermal power of the pellet heating. To me it seems that pellet_minimum_power = pellet_maximum_power * 1/12 can be replaced by pellet_minium_power = building_heating_load * 1/12 (as pellet_maximum_power is sized by building_heating_load)
- B8: yes agree, but do you know already what types will be used for Catalogue? In general, I would try to reduce the number of "Any" types in the code. 
- B9: I have no strong opinion. I guess it could make sense to use as presets those config values and files whoch are stored hisim/inputs in the repo and which are often used for testing 
- B10: agree
- B11: not sure. right now the HDS is global but in the future we might have several HDS systems per building (in MFHs)
- C10: agree, unless this change is very unconvenient for the other users
- C11: Making the choice of HDS system automatical could be helpful but where would you draw the line? meaning for what buildings or building ages would you propose conventional radiators, low temperature radiators and floorheating, on what "source" would you base this decision?
- D13: I thought most of these files were moved already to obsolete (see in main branch). if not, they can be moved to obsolete
- Q5: agree

k-rieck:
some other general comments:

- my first feeling is that in general this whole sizing module is quite complex and has several layers of abstraction whereas the actual "sizing" logic that we currently use in HiSim is actually pretty simple and restricted to a handful of use cases
- that's why my question is: is this module the best and simplest architecture for our requirements and is it justified? does claude know what use cases for sizing we currently have? 
- for example: do we need all of these SizingLaws? 
-  class _FactTerm(SizingLaw):
        - class _ScaledLaw(SizingLaw):
        - class _ClampedLaw(SizingLaw):
        - class _RoundedLaw(SizingLaw):
        - class _ConstantLaw(SizingLaw):
        - class _FunctionLaw(SizingLaw):
-

k-rieck (approved):

PR #594, opened 2026-08-27
Renovisor backend

noah80:
Revised Renovisor Translation layer spec


energy_systems/README.md
valentinjanser, line 9: I would first introduce the two kinds of files (energy system and simulation) and then show/explain the individual configs).

valentinjanser, line 41: "every connection between the components of the simulated scenario"

PR #595, opened 2026-08-27
Revised RenoVisor translation layer: requirements for the v2 measure catalogue

noah80:
## Problem

RenoVisor's measure catalogue arrived (33 measures, 33 options) and it replaces the
discriminated-union measure model this branch had proposed in `mockups/measures.schema.yaml`.
The two differ in substance, not spelling: the catalogue states a material and a thickness
where the old model stated a resulting U-value, sizes PV and battery as "percent of roof
area" and "days to cover", and lets several measures write the same U-value field. None of
that is covered by `requirements.md`, and the catalogue itself was not in the repository.

## Done

- `roadmap/renovisor/mockups/measures.yaml` — the catalogue, checked in verbatim so the
  requirements have a fixed text to refer to.
- `roadmap/renovisor/measures_v2_requirements.md` — what the catalogue changes for HiSim,
  written as a delta on `requirements.md` (its R/A/Q ids are reused). Every code-facing
  statement was checked against `4612b899`:
  - A7 is reversed: HiSim must derive U-values from material + thickness, and needs a
    conductivity table, default thicknesses (the thickness option is Experts-only) and a
    formula. 15 of 33 measures depend on it.
  - Insulation layers on one element must add up, not overwrite (amends A6.3).
  - PV and battery sizes need named sizing laws.
  - The format's `variants:` block already covers the solar-thermal, DHW and EMS
    choices, so one base file per heat generator (11) is enough. Q4 is answered; R4 gains
    variant selection as a permitted write.
  - Counted against HiSim today: 2 measures work, 15 need the derivation, 7 wait for P4,
    9 have no model — 4 of those need only a new `BuildingConfig` override field each.
  - A proposal for the measure→HiSim mapping as three data tables (measure effects,
    physics lookups, inventory→energy-system bindings) with seven CI checks. The first two
    tables never name a HiSim component, so a P4 rename touches one row of the third.
  - 12 requirements on HiSim, 8 on the catalogue/contract, 4 open questions with
    recommendations, and an order of work that starts with what needs neither P4 nor the
    frontend.

## Result

The branch now carries the catalogue and a reviewable statement of what it costs. Two inputs
are still missing from the repository and are marked as blocking: the `openapi.yaml`
revision the catalogue refers to (the copy here is v0.3.0-draft of 2026-08-27 and lacks
`worst_window_glazing_panes`, `hvo_heating`, the snake_case enum values and the
`ventilation_system` block) and `dependencies.yaml`. Nothing under `hisim/` changes.

### DIFF
--- a/roadmap/renovisor/field_inventory.md
+++ b/roadmap/renovisor/field_inventory.md
@@ -33,7 +33,7 @@ This file holds the raw survey. The requirements document quotes only the headli
 | 8 | `…general.number_of_apartments` | used | `BuildingConfig.number_of_apartments` |
 | 9 | `…general.set_heating_temperature_in_celsius` | used | same name |
 | 10 | `…general.set_cooling_temperature_in_celsius` | used | same name |
-| 11 | `…general.retrofit_status` | used | TABULA variant `.001` / `.002` / `.003` |
+| 11 | `…general.retrofit_status` | used | TABULA variant `.001` / `.002` / `.003` of the generic `Gen.ReEx.001` rows. Higher trailing numbers in the dataset (e.g. `IE.N.SFH.01.325SB…`) are special sub-typologies, not further refurbishment states; the translation excludes them. |
 | 12 | `…general.max_thermal_building_demand_in_watt` | used | same name; precedence against the sizing kernel undefined (requirement A11) |
 | 13–22 | `…envelope_details.*` (10 fields) | used | `BuildingConfig.{floor,facade,roof,window,door}_{u_value_in_watt_per_m2_per_kelvin,area_in_m2}` — **identical names** |
 | 23 | `occupancy_config.residents_count` | approximated | one of 66 `utspclient.helpers.lpgdata.Households` entries |
@@ -93,6 +93,14 @@ consumption but no mileage. HiSim's driving distance comes from the LPG travel-r
 occupancy profile, so mileage is *implied* by field 26 and cannot be stated independently.
 Consequence: a car's annual energy is a function of `travel_route_set`, which the contract
 documents as an occupancy field, not a vehicle field.
+
+**Decided (2026-09-05, review):** the next contract revision reinstates `vehicles[].km_per_year`
+as an optional field. The travel-route set encodes the commute, and commute distance under-states
+a household whose kilometres are holidays and hobbies — a car can do 20 000 km/year on a 5 km
+commute without any of it touching the route set. When stated, `km_per_year` scales the car's
+annual driving energy; the travel-route set remains the fallback and still shapes *when* the car
+is away from home. This puts a requirement on the translation layer: a stated mileage must scale
+car consumption independently of the LPG profile (A5).
 
 ## 3. `Kpis` — 13 leaf fields
 
--- a/roadmap/renovisor/field_inventory.md
+++ b/roadmap/renovisor/field_inventory.md
@@ -1,0 +1,190 @@
+# RenoVisor v0.3 contract ↔ HiSim — counted field inventory
+
+**Companion to** `roadmap/renovisor/requirements.md` · **Date:** 2026-08-27
+**Sources surveyed:** `roadmap/renovisor/openapi.yaml` (v0.3.0-draft), `hisim/components/`,
+`hisim/economics/`, `hisim/energy_system/`, `hisim/postprocessing/kpi_computation/`,
+`energy_systems/gas_boiler_household.energy_system.yaml`,
+`roadmap/declarative_energy_systems/plan.md`.
+
+This file holds the raw survey. The requirements document quotes only the headline numbers.
+
+## 0. Status vocabulary
+
+| Status | Meaning |
+|---|---|
+| `used` | A HiSim configuration field or energy-system file element consumes the value directly. |
+| `approximated` | Consumed, but the contract's value space is larger than HiSim's; the mapping is lossy and many-to-one. |
+| `blocked` | A HiSim field exists that would consume it, but the owning class has no preset/constructor yet, so no energy-system file can carry it (P4 sweep). |
+| `structural` | Not a configuration value at all: it decides which components exist and how they are wired. |
+| `no consumer` | Nothing in HiSim consumes it and nothing is planned to. |
+| `non-simulation` | Legitimate input, but for the cost / scheduling / grant layer, not for the physics. |
+
+## 1. `HomeInventoryInput` — 79 leaf fields
+
+| # | Field | Status | HiSim counterpart / note |
+|---|---|---|---|
+| 1 | `location.country_code` | used | `weather.LocationEnum`; `IE = ("Dublin", "NSRDB_15min", …, "165308_53.37_-6.26_2019.csv")`. Also the TABULA country prefix. |
+| 2 | `location.region` | no consumer | One weather station per country code. 44 `LocationEnum` entries, 1 of them Irish. |
+| 3 | `location.eircode_or_postcode` | no consumer | Not used for weather, TABULA or tariffs. |
+| 4 | `building_config.general.tabula_building_type` | used | TABULA code segment. IE rows: 48 `SFH`, 51 `TH`, 19 `AB`, 6 hybrid; **0 `MFH`** (124 rows total). |
+| 5 | `…general.building_heat_capacity_class` | used | `BuildingConfig.building_heat_capacity_class` |
+| 6 | `…general.construction_year` | used | TABULA age band (`Year1_Building`–`Year2_Building`) |
+| 7 | `…general.conditioned_floor_area_m2` | used | `BuildingConfig.absolute_conditioned_floor_area_in_m2` |
+| 8 | `…general.number_of_apartments` | used | `BuildingConfig.number_of_apartments` |
+| 9 | `…general.set_heating_temperature_in_celsius` | used | same name |
+| 10 | `…general.set_cooling_temperature_in_celsius` | used | same name |
+| 11 | `…general.retrofit_status` | used | TABULA variant `.001` / `.002` / `.003` |
+| 12 | `…general.max_thermal_building_demand_in_watt` | used | same name; precedence against the sizing kernel undefined (requirement A11) |
+| 13–22 | `…envelope_details.*` (10 fields) | used | `BuildingConfig.{floor,facade,roof,window,door}_{u_value_in_watt_per_m2_per_kelvin,area_in_m2}` — **identical names** |
+| 23 | `occupancy_config.residents_count` | approximated | one of 66 `utspclient.helpers.lpgdata.Households` entries |
+| 24 | `occupancy_config.residents_type[]` | approximated | same; `{adult, kid}` does not address the catalogue |
+| 25 | `occupancy_config.residents_employment_status[]` | approximated | same; 5 statuses × n residents ≫ 66 catalogue entries |
+| 26 | `occupancy_config.travel_route_set` | used | `TravelRouteSets`, exactly the 6 values the contract lists |
+| 27 | `energy_system_config.heating_system.system` | used | `hisim.loadtypes.HeatingSystems` — the contract's 10 values *are* the enum's 10 values |
+| 28 | `…heating_system.heatpump_flow_temperature_in_celsius` | blocked | `more_advanced_heat_pump_hplib` unconverted; the real knob sits on the HDS controller |
+| 29 | `…heating_system.power_in_watt` | used | `GenericBoilerConfig` power; precedence against `AUTO` undefined (A11) |
+| 30 | `…heating_system.installation_year` | non-simulation | no ageing model; cost/replacement input |
+| 31 | `…heating_system.with_dhw_preparation` | blocked | needs `simple_water_storage.SimpleDHWStorage` (unconverted) |
+| 32 | `…heating_system.heat_distribution_system` | used | `lt.HeatDistributionSystemType`: `"Floorheating"`, `"Conventional Radiator"`, `"Low Temperature Radiator"` — exact string match |
+| 33 | `…water_storage.hot_water_storage.volume_in_liters` | blocked | `SimpleHotWaterStorage` unconverted |
+| 34 | `…hot_water_storage.installation_year` | non-simulation | |
+| 35 | `…domestic_hot_water_storage.volume_in_liters` | blocked | `SimpleDHWStorage` unconverted |
+| 36 | `…domestic_hot_water_storage.installation_year` | non-simulation | |
+| 37 | `…photovoltaics.power_in_watt` | blocked | `PVSystemConfig.power_in_watt` (module unconverted) |
+| 38 | `…photovoltaics.azimuth_in_degree` | blocked | `PVSystemConfig.azimuth` |
+| 39 | `…photovoltaics.tilt_in_degree` | blocked | `PVSystemConfig.tilt` |
+| 40 | `…photovoltaics.installation_year` | non-simulation | |
+| 41 | `…photovoltaics.remaining_performance_in_percent` | no consumer | no degradation model; `share_of_maximum_pv_potential` is a *sizing* share, not a derate |
+| 42 | `…battery_storage.capacity_in_kwh` | blocked | `BatteryConfig.custom_battery_capacity_generic_in_kilowatt_hour` |
+| 43 | `…battery_storage.power_in_watt` | no consumer | only `custom_pv_inverter_power_generic_in_watt` (the inverter) exists |
+| 44 | `…battery_storage.installation_year` | non-simulation | |
+| 45 | `…battery_storage.state_of_health_in_percent` | no consumer | no capacity-fade model |
+| 46 | `…solar_thermal_system.collector_area_in_m2` | blocked | `SolarThermalSystemConfig.area_m2` — **name mismatch** |
+| 47 | `…solar_thermal_system.azimuth_in_degree` | blocked | `.azimuth` |
+| 48 | `…solar_thermal_system.tilt_in_degree` | blocked | `.tilt` |
+| 49 | `…solar_thermal_system.used_for_space_heating` | structural | decides which storage the collector feeds |
+| 50 | `…solar_thermal_system.used_for_dhw` | structural | same |
+| 51 | `…solar_thermal_system.installation_year` | non-simulation | |
+| 52 | `…electric_vehicles[].model` | no consumer | |
+| 53 | `…electric_vehicles[].consumption_in_kwh_per_km` | blocked | `CarConfig.consumption_per_km` |
+| 54 | `…electric_vehicles[].battery_capacity_in_kwh` | blocked | `advanced_ev_battery_bslib` |
+| 55 | `…electric_vehicles[].construction_year` | non-simulation | |
+| 56 | `…electric_vehicles[].charging_location` | no consumer | the setups charge at home only |
+| 57 | `…fossil_vehicles[].fuel_source` | blocked | `CarConfig.fuel` (`lt.LoadTypes`) |
+| 58 | `…fossil_vehicles[].model` | no consumer | |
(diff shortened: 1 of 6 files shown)


roadmap/renovisor/mockups/heat_pump_household.energy_system.yaml
```diff
@@ -0,0 +1,231 @@
...
+  home inventory 3f9a2c… by the RenoVisor translation layer; do not hand-edit.
+
+components:
+
+  weather:
+    class: hisim.components.weather.Weather
+    constructor:
+      for_location:
+        location: IE                       # [override] location.country_code
+        data_source: NSRDB_15MIN           # [base] the only Irish dataset (2019)
```
valentinjanser: This implies that all inventories choosing "HeatPump" are located in Ireland? (Or at least, the NSRDB_15MIN is fitting. Is that the case?)

roadmap/renovisor/mockups/measures.schema.yaml
```diff
@@ -0,0 +1,488 @@
...
+    - $ref: '#/components/schemas/ReplaceHeatGeneratorMeasure'
+    - $ref: '#/components/schemas/ReplaceHeatDistributionMeasure'
+    - $ref: '#/components/schemas/InstallThermalStorageMeasure'
+    - $ref: '#/components/schemas/InstallPhotovoltaicsMeasure'
+    - $ref: '#/components/schemas/InstallBatteryStorageMeasure'
+    - $ref: '#/components/schemas/InstallSolarThermalMeasure'
+    - $ref: '#/components/schemas/InstallVentilationMeasure'
+    - $ref: '#/components/schemas/ReplaceVehicleMeasure'
+    - $ref: '#/components/schemas/InstallChargerMeasure'
+    - $ref: '#/components/schemas/DecommissionMeasure'
```
valentinjanser, line 59: These names should later be adapted to the measure database that will be located in the same repo as the openapi.yaml definition this will be part of.

roadmap/renovisor/mockups/measures.schema.yaml
```diff
@@ -0,0 +1,488 @@
...
+        element:
+          allOf: [{ $ref: '#/components/schemas/BuildingElement' }]
+          description: One of roof, facade, floor. Rejected for window and door, which are replaced, not insulated.
+        material_id:
+          type: string
+          description: A `/materials` entry. Optional, and PRICING-ONLY unless the catalogue carries thermal conductivity (A7).
+        thickness_mm:
+          type: number
+          description: Optional, pricing-only. It does not determine the U-value; the U-value is stated outright.
+        resulting_u_value_in_watt_per_m2_per_kelvin:
```
valentinjanser, line 141: Does this mean that your idea is to have the frontend compute the resulting U-values? 
I ask because with the current plans as I understood them, the frontend would never have to provide/receive any U-value (but COULD provide the current one)...

roadmap/renovisor/mockups/measures.schema.yaml
```diff
@@ -0,0 +1,488 @@
...
+  description: |
+    Names one vehicle of the inventory. Requires `HomeInventoryInput`'s vehicle
+    entries to carry a stable `vehicle_id`, which they do not today (footer F4).
+    Positional references were rejected: an index is not stable under a list
+    reorder, and a package's hash must not change because the inventory was
+    re-serialised.
+  type: object
+  required: [vehicle_id]
+  properties:
+    vehicle_id: { type: string, examples: [car-1] }
```
valentinjanser, line 352: This is the schema used to specify the fossil fuel car that is being replaced. Do we not want to know what the fuel consumption was (e.g., for the cost module)? And if we have all this data, maybe use the VehicleRef for both target & replacement?

roadmap/renovisor/mockups/measures.schema.yaml
```diff
@@ -0,0 +1,488 @@
...
+  type: object
+  properties:
+    preferred_start: { type: string, examples: ['2026-09'] }
+    phases:
+      type: array
+      items:
+        type: object
+        required: [measure_ids]
+        properties:
+          starts_in_year: { type: integer, examples: [2026] }
```
valentinjanser: If the preferred_start is given as yyyy-MM, we should maybe do it here as well.

roadmap/renovisor/mockups/measures.schema.yaml
```diff
@@ -0,0 +1,488 @@
...
+#     conforming client could have written a valid schedule.
+#
+# F3  `measure_id` must be derived from content, not assigned. Two users
+#     specifying the same renovation must produce the same `pkg_id`, and the
+#     contract's dedup promise ("submitting identical blocks returns the same
+#     pkg_id") fails outright if a client is free to mint random ids.
+#
+# F4  Vehicle entries in `HomeInventoryInput` need a stable `vehicle_id` so a
+#     measure can name one. This is a change to the inventory, not only to the
+#     measure — the two cannot be revised separately.
```
valentinjanser: As far as I see, we currently could simply count the number of fossil fuel-powered cars left in the "energy system".

roadmap/renovisor/mockups/measures.schema.yaml
```diff
@@ -0,0 +1,488 @@
...
+#     `roof_covering` all fall into `roof`, and `interior_walls` maps to nothing
+#     thermal at all. Measures target the five, because those are what the
+#     thermal model resolves. Whether the condition vocabulary shrinks to match,
+#     or stays as a finer-grained survey instrument that maps onto the five, is
+#     a decision for the frontend team (A21).
+#
+# F6  `install_ventilation_system` has nowhere to write: `condition_assessment`
+#     can report a ventilation system's condition, but `energy_system_config`
+#     has no ventilation block, so the inventory cannot say one exists. Either
+#     the block is added or the measure goes.
```
valentinjanser: The block has to be added, same for the air conditioning.

roadmap/renovisor/field_inventory.md
```diff
@@ -0,0 +1,198 @@
...
+| 2 | `location.region` | no consumer | One weather station per country code. 44 `LocationEnum` entries, 1 of them Irish. |
+| 3 | `location.eircode_or_postcode` | no consumer | Not used for weather, TABULA or tariffs. |
+| 4 | `building_config.general.tabula_building_type` | used | TABULA code segment. IE rows: 48 `SFH`, 51 `TH`, 19 `AB`, 6 hybrid; **0 `MFH`** (124 rows total). |
+| 5 | `…general.building_heat_capacity_class` | used | `BuildingConfig.building_heat_capacity_class` |
+| 6 | `…general.construction_year` | used | TABULA age band (`Year1_Building`–`Year2_Building`) |
+| 7 | `…general.conditioned_floor_area_m2` | used | `BuildingConfig.absolute_conditioned_floor_area_in_m2` |
+| 8 | `…general.number_of_apartments` | used | `BuildingConfig.number_of_apartments` |
+| 9 | `…general.set_heating_temperature_in_celsius` | used | same name |
+| 10 | `…general.set_cooling_temperature_in_celsius` | used | same name |
+| 11 | `…general.retrofit_status` | used | TABULA variant `.001` / `.002` / `.003` |
```
valentinjanser: Watch out, there are also numbers larger than 003. I do not currently know their meaning.

roadmap/renovisor/field_inventory.md
```diff
@@ -0,0 +1,198 @@
...
+decisions with no base file yet.
+
+## 2. Annual mileage: a field the contract dropped
+
+The v1 contract carried `vehicles[].kmPerYear`
+(`hisim/renovisor/examples/example_1_gas_to_heatpump_pv_insulation.json`). v0.3 carries per-km
+consumption but no mileage. HiSim's driving distance comes from the LPG travel-route set and the
+occupancy profile, so mileage is *implied* by field 26 and cannot be stated independently.
+Consequence: a car's annual energy is a function of `travel_route_set`, which the contract
+documents as an occupancy field, not a vehicle field.
```
valentinjanser, line 95: I still wonder whether this is the better way of modelling... If you drive a lot of km/year, but it is due to holidays, this doesn't really impact your household energy situation. However, km/year is maybe more revealing than the commute distance (to work as I understand?). Even if the latter is 5 km, you can easily drive 20000 km/year for hobbies etc.

roadmap/renovisor/requirements.discussion.md
```diff
@@ -0,0 +1,410 @@
...
+unchanged hash. The contract has no version field anywhere.
+
+*Options.* (a) **Version in the hash** — `hi_id` and `pkg_id` stay input-only, but the result
+URL resolves per version. Consequence: a model update invalidates every cached result, which is
+correct and expensive. (b) **Version namespaces the result store**, with the URL unchanged and
+an internal generation counter. Consequence: cheap for the frontend, but a client that cached a
+year-old response still shows stale numbers. (c) **Weaken `immutable`** to a short max-age plus
+an ETag. Consequence: correctness at the cost of the contract's central caching claim.
+
+*Recommendation.* (a) for correctness, with the version exposed in the finished result so a
```
valentinjanser: I agree. It also sounds logical that if we change the model, we have to recompute.
(b) will likely cause massive headaches
(c) is basically the same as (a), but with less guarantees. I do not see an advantage.

valentinjanser (approved):

roadmap/renovisor/openapi.yaml
```diff
@@ -0,0 +1,948 @@
...
+      type: object
+      properties:
+        measures:
+          type: array
+          items: { $ref: '#/components/schemas/Measure' }
+        grants: { $ref: '#/components/schemas/Grants' }
+        schedule: { $ref: '#/components/schemas/Schedule' }
+        financing: { $ref: '#/components/schemas/Financing' }
+
+    Measure:
```
k-rieck, line 829: now you have one Measure schema here and one (correct one) defined in mockups. Should be consistent

roadmap/renovisor/mockups/measures.schema.yaml
```diff
@@ -0,0 +1,488 @@
...
+
+BuildingElement:
+  x-contract-status: provisional
+  description: |
+    The elements the thermal model actually resolves, and the ONLY vocabulary a
+    measure may target. It is deliberately the five of `envelope_details`, not
+    the nine of `condition_assessment.building_elements`; see footer F5, which
+    is a defect in the draft this mockup does not paper over.
+  type: string
+  enum: [roof, facade, floor, window, door]
```
k-rieck, line 116: in Cristinas db there is also basement ceiling, basement floor and walls and ground-contacting components

roadmap/renovisor/mockups/measures.schema.yaml
```diff
@@ -0,0 +1,488 @@
...
+    - type: object
+      required: [element, resulting_u_value_in_watt_per_m2_per_kelvin]
+      properties:
+        element:
+          allOf: [{ $ref: '#/components/schemas/BuildingElement' }]
+          description: One of roof, facade, floor. Rejected for window and door, which are replaced, not insulated.
+        material_id:
+          type: string
+          description: A `/materials` entry. Optional, and PRICING-ONLY unless the catalogue carries thermal conductivity (A7).
+        thickness_mm:
```
k-rieck, line 138: there is Material but also Format like rigid board, semi-flexible mats, blown-in insulation, flexible insulation blankets, bale (straw), insulating plaster etc. (see Cristinas db)

roadmap/renovisor/mockups/measures.schema.yaml
k-rieck: in Cristinas db (see sheet "Tabula Analyse IE"), the thickness of the new insulation is calculated based on the existing U-value (from houseinventory or Tabula type) and the target U-value for the given element (based on regulations, see sheet "insulation measures"). so its rather "resulting thickness" than resulting "U-value"

roadmap/renovisor/mockups/measures.schema.yaml
```diff
@@ -0,0 +1,488 @@
...
+ReplaceWindowsMeasure:
+  x-contract-status: provisional
+  description: Replace the glazing. Area is unchanged unless stated.
+  x-writes: building_config.envelope_details.window_u_value_in_watt_per_m2_per_kelvin (and window_area_in_m2 when given)
+  allOf:
+    - $ref: '#/components/schemas/MeasureBase'
+    - type: object
+      required: [resulting_u_value_in_watt_per_m2_per_kelvin]
+      properties:
+        resulting_u_value_in_watt_per_m2_per_kelvin: { type: number, examples: [0.9] }
```
k-rieck: also here each countries seem to have "target U-values" for windows (see Cristinas db, sheet "insulation measures")

roadmap/renovisor/mockups/measures.schema.yaml
```diff
@@ -0,0 +1,488 @@
...
+
+ReplaceDoorMeasure:
+  x-contract-status: provisional
+  x-writes: building_config.envelope_details.door_u_value_in_watt_per_m2_per_kelvin
+  allOf:
+    - $ref: '#/components/schemas/MeasureBase'
+    - type: object
+      required: [resulting_u_value_in_watt_per_m2_per_kelvin]
+      properties:
+        resulting_u_value_in_watt_per_m2_per_kelvin: { type: number, examples: [1.2] }
```
k-rieck: same comments as above

roadmap/renovisor/mockups/measures.schema.yaml
```diff
@@ -0,0 +1,488 @@
...
+        material_id: { type: string, description: Pricing only. }
+
+ImproveAirtightnessMeasure:
```
k-rieck, line 186: would that be part of "Shallow Measures"?

roadmap/renovisor/mockups/measures.schema.yaml
```diff
@@ -0,0 +1,488 @@
...
+    8 kW says 8000 and not 5000.
+  x-writes: energy_system_config.photovoltaics
+  allOf:
+    - $ref: '#/components/schemas/MeasureBase'
+    - type: object
+      required: [power_in_watt]
+      properties:
+        power_in_watt: { type: number, examples: [5000] }
+        azimuth_in_degree: { type: number, examples: [180] }
+        tilt_in_degree: { type: number, examples: [35] }
```
k-rieck: that will be determined by building roof shape and orientation or would you like allow all values?

k-rieck (approved):

roadmap/renovisor/mockups/heat_pump_household.energy_system.yaml
noah80: No — the dataset follows the location, never the measure: 'IE' is an override from
  location.country_code, and NSRDB_15MIN is simply the one Irish dataset. Ireland is the product's current scope. Clarified the comment in
  the mockup.

roadmap/renovisor/mockups/measures.schema.yaml
noah80: Agreed — recorded as footer F9: the variant shapes are the proposal, the type spellings await
  reconciliation with the measures database once it lands beside openapi.yaml.

roadmap/renovisor/mockups/measures.schema.yaml
noah80: Katharina's description matches the intent: the regulatory target U-value per element per country
  IS resulting_u_value, and the thickness is what's derived from it — for pricing. The simulation consumes exactly one authoritative
  U-value; nobody computes U-values in the frontend. The descriptions and footer F8 now say this explicitly.

roadmap/renovisor/mockups/measures.schema.yaml
noah80: The replaced car's consumption isn't restated because 'target' points into the inventory, which
  already carries it — that's where the cost module reads it. VehicleRef can't describe the replacement: it's a new vehicle the inventory
  has never seen, so it needs a specification, not a name. Added to the description.

roadmap/renovisor/mockups/measures.schema.yaml
noah80: Done — phases now carry starts_in as yyyy-MM, same convention as preferred_start.

roadmap/renovisor/mockups/measures.schema.yaml
noah80: Considered and recorded in F4: counting is ambiguous the moment a household has two fossil cars
  with different consumption — which one goes changes the result, which the content-hash promise can't tolerate. So VehicleRef stays.

roadmap/renovisor/mockups/measures.schema.yaml
noah80: Agreed and recorded as the F6 decision: energy_system_config gains a ventilation block and an
  air-conditioning block, and the measure stays.

roadmap/renovisor/field_inventory.md
noah80: Those are the special sub-typologies (e.g. IE.N.SFH.01.325SB…), not further refurbishment states —
  the translation uses only the generic Gen.ReEx.001 rows and excludes the rest. Now stated in row 11.

roadmap/renovisor/field_inventory.md
noah80: Agreed — commute distance under-states hobby/holiday mileage. Decided: the next contract revision
  reinstates an optional vehicles[].km_per_year that scales the car's annual driving energy, with travel_route_set as the fallback (A5 is
  now decided; field_inventory §2 has the details).

roadmap/renovisor/requirements.discussion.md
noah80: Adopted as the decision: (a), version in the hash, with the version exposed in the finished result.
  Q2 is marked decided and R14/A15/AC10 proceed on that basis.

roadmap/renovisor/openapi.yaml
noah80: Fixed — the flat Measure (and the phase_order Schedule) in openapi.yaml are now marked superseded
  with pointers to the mockup, kept only as stubs so the file's refs stay resolvable until the mockup is pasted in.

roadmap/renovisor/mockups/measures.schema.yaml
noah80: Decided: the measure vocabulary stays the five elements the thermal model can write, and F5 now
  carries the mapping — basement ceiling/floor and ground-contacting components fold into 'floor', above-ground walls into 'facade'. The
  finer survey/database categories stay as instruments that map onto the five.

roadmap/renovisor/mockups/measures.schema.yaml
noah80: Format lives in the /materials catalogue row that material_id points at (rigid board, blown-in,
  blankets, bale, plaster…), so the measure doesn't repeat it — now stated in the field's description.

roadmap/renovisor/mockups/measures.schema.yaml
noah80: Yes — the window description now says the value is typically the country's regulatory target
  U-value from the measures database.

roadmap/renovisor/mockups/measures.schema.yaml
noah80: Same as for windows — noted on the door measure too.

roadmap/renovisor/mockups/measures.schema.yaml
noah80: Yes — labelled as a shallow measure (the database's taxonomy) in the description.

roadmap/renovisor/mockups/measures.schema.yaml
noah80: Free values, deliberately: the inventory carries no roof geometry, so the backend can't validate
  against the roof — whoever builds the request derives azimuth/tilt from it, and omitted values fall back to the base file's defaults. Now
  documented on the fields.

PR #613, opened 2026-09-01
D-23: build the car from its configuration

noah80:
Converts `Car` to build from its own configuration, split out of the P3 branch (#598) so it can be
reviewed as what it is: a change to a shipped component, its wiring and the v1 JSON path.

Answers **D-23** — convert the car now, around `SizingContext` rather than through it.

## Why it was stuck

`Car.__init__` required a `data_dict_with_car_information` positional argument that only existed at
runtime, assembled by the UTSP connector. Nothing could build a car from configuration alone, so the
v1 JSON path carried two special cases for it and the declarative path could not express it at all
(`EF-33`).

The driving profile now travels through the **simulation repository** rather than a constructor
argument. `SimRepository`, deliberately, not the singleton — that scales to districts and multi-unit
buildings, and the singleton is on its way out (#601).

## What changed

| Commit | |
|---|---|
| Hand the driving profile over through the simulation repository | new `lpg_car_information.py`; the connector publishes |
| Build the car from its configuration | `generic_car.py` rewritten, plus 9 tests where there were none |
| Retire the v1 JSON path's two special cases | `json_executor.py`, `json_generator.py` — 40 lines lighter |
| Build each car of the sizer from a configuration of its own | the setup |
| Let the energy manager find its car on either wiring path | see below |
| Answer D-23 | roadmap |
| Let the wire format admit a class that has no preset at all | v3 schema |
| Carry the cache-metadata call across the car's rewrite | see below |
| Re-record the car setup's v1 scenario JSON | regenerated, verified byte-identical |

**The energy manager change is worth a look.** It found each steered participant by looking for the
class name inside the name of the output that steers it. That works for outputs it names itself, and
not for a dispatch target, which is named by whoever wired the participant in — a Python setup passes
a class-derived prefix, an energy-system file derives `DispatchTo<instance>_<input>`. So one system
reported the car's grid consumption when built from the setup and omitted it when built from the
recorded file: the KPI layer disagreeing with itself about a system whose every result column was
identical. It now matches on the tags the dispatch output carries, which both paths agree on.

## The split found a bug that neither branch had

The car conversion predates the cache-metadata file every entry now gets (#608 F5). On the P3 branch
the two never met. Cherry-picked onto today's `main` they merged **with no textual conflict** and
produced code that raised on the first car built:

```python
utils.build_cache_key_string(config, self.my_simulation_parameters)
#                            ^^^^^^ NameError: the rewrite moved this into a method
#                                   where that local no longer exists
```

Fixed to pass `self.config` — the same object the cache key three lines above is derived from, which
is what makes the metadata describe the entry it sits beside rather than something else.

This would have reached `main` the day P3 merged, and nothing would have connected it to #608. Two
branches that are each green can still be broken together, and only putting one on top of the other
finds it.

## What deliberately stayed with P3

- the `energy_systems/*.energy_system.yaml` twin of this setup — P3 owns the recorder that writes it
  and the freshness gate that keeps it honest
- the parity rig's renaming-table entries (`scripts/p3_parity_renamings.py`)

This branch carries only what `main`'s own gates already cover.

## Verification

- `tests/test_generic_car.py`: 9 passed (new — the component had no tests)
- scenario-JSON freshness, whole fleet: 22 OK, 0 FAILED, no drift
- golden-check and golden-json-check on an EMS-bearing pair: both OK
- mypy 282 / 23 / 184 clean; pylint 10.00/10; prospector 0 on `hisim/` and `system_setups/`, 26 on
  `tests/` (baseline)
- `pytest -m base`: 4508 passed, 1 skipped, 0 failed


hisim/components/generic_car.py
```diff
@@ -40,115 +40,35 @@
...
-            # get car names
-            car_name = (
-                car_location["LoadTypeName"]
-                .split(" - ")[1]
-                .translate(str.maketrans({" ": "_", ",": "", "/": "", ".": ""}))
-            )
-            car_names.append(car_name)
+    #: Consumption per kilometre in kWh/km, CO2 footprint of manufacture in kg and investment cost
+    #: in Euro of the electric vehicle :meth:`for_household` configures for an electric car.
+    ELECTRIC_VEHICLE: ClassVar[Tuple[float, float, float]] = (0.15, 8899.4, 44498.0)
```
k-rieck, line 55: is there a way to have values and parameter names and units connected, like electricity_consumption_in_kwh_per_hour = 0.15 etc.? having the description in the comments only, is not ideal in my opinion

k-rieck: nevermind. I see that this is added in line 134

k-rieck (approved):

PR #614, opened 2026-09-01
Let an undefined ratio be undefined, so a household with no electricity can report

noah80:
simple_air_conditioner_household_building_sizer builds a weather, a building and an air conditioner: no occupancy, no generation, no meter. Its total electricity
  consumption is genuinely zero, and the KPI layer divided by it (ZeroDivisionError at kpi_preparation.py:560), then multiplied a self-sufficiency rate that is None for want of a
  meter (TypeError at :596) -- the same line that stopped the two setups before it, which were fixed by giving them meters. This one cannot be: there is no electricity to meter.
  Both places now emit the entry without a value, which is what the same module already does for a self-consumption rate with no production. A ratio of something to nothing is
  undefined, not zero. Eight of the setup's 68 KPIs are now empty, all of them proportions of a consumption that does not exist, and the new test names five and asserts they stay
  empty -- the failure mode of this fix is somebody later deciding the blanks look untidy. The file's three existing tests build the component graph and never run the setup or
  ask for KPIs, which is why it passed its own suite while broken.

### DIFF
--- a/hisim/postprocessing/kpi_computation/kpi_preparation.py
+++ b/hisim/postprocessing/kpi_computation/kpi_preparation.py
@@ -556,8 +556,21 @@ class KpiPreparation:
         """Compute the ratio of two values.
 
         ratio = denominator / numerator * 100 [%].
+
+        Every caller divides by a total consumption, and a system can legitimately consume nothing:
+        the simple air-conditioner household has no occupancy and no generation at all, so it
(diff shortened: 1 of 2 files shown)


k-rieck (approved):
general comment:
- yes, this fixes the simple air conditioned house. but we dont want to keep it that way right? We should raise an issue saying "simple air conditioned house needs electricity meter", would you agree? then this test would become obsolete later on

PR #622, opened 2026-09-02
Give the parallel drivers one place to borrow local-LPG calculation indices from

noah80:
## Problem

pylpg computes inside its installed package in a directory named after the calculation index, so a driver running several HiSim processes must hand each worker slot its own index or they corrupt each other's sqlite files. `regenerate_scenario_jsons.py` grew an inline queue for this. The P3 recorder was expected to need the same, and the environment variable's name was spelled independently in each driver — a disagreement nothing would report, since a child reading a name nobody sets falls back to its pid and works until two collide.

## What was done

- `LpgBaseIndexPool` in `hisim/components/pylpg_workspace.py`, next to the index arithmetic it belongs to: lends an index for the length of a block, returns it in a `finally` (a child that raised or was killed still frees its slot), refuses a zero-slot pool, and `child_environment()` builds a child's environment with the variable set, so the name lives in one place.
- The regenerator borrows from it instead of its inline queue. Reaching the module from the driver needs the checkout on `sys.path` first (same defect as #621, one process further up) — demonstrated, not asserted.
- Six tests, including a round trip through `PylpgWorkspace.default_base_index()` so the two halves of the contract cannot drift.

## Result

One implementation of index lending, shared name, tested. Second commit: docstrings rewritten in plain language after review feedback (what → example → short why), `announced_in` renamed `child_environment`. **Honest caveat:** the P3 recorder turned out not to need it — it is sequential by design (two setups needing the same new parameters file must share it), and #598 un-pins its index instead. So this PR is a clean-up with tests rather than a prerequisite; merge or close on that basis. Based on #621 (both edit `regenerate_one`); GitHub retargets to `main` when #621 merges.

### DIFF
--- a/hisim/components/pylpg_workspace.py
+++ b/hisim/components/pylpg_workspace.py
@@ -14,6 +14,12 @@ is claimed before it is created and released once the attempt ends, whether it s
 claiming one that already exists fails immediately with a message naming the index, because the
 alternative is two runs silently interleaving in one folder. See ``roadmap/pylpg_flakiness.md`` F3
 and F4.
+
+A driver that runs several HiSim processes at once needs one more thing: a way to give each of its
+worker slots an index of its own and take it back when the slot is reused. That is
+:class:`LpgBaseIndexPool`, which lives here rather than in any one driver so that the two scripts
+which need it -- the scenario-JSON regenerator and the energy-system recorder -- cannot drift apart
+on how indices are handed out or on what the environment variable is called.
 """
 
 # clean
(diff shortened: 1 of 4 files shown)


hisim/components/pylpg_workspace.py
```diff
@@ -401,3 +408,80 @@ def release(cls, calculation_indices: List[int]) -> None:
...
+
+
+class LpgBaseIndexPool:
+    """Hands out distinct local-LoadProfileGenerator base indices to the slots of a parallel driver.
+
+    A driver that runs several HiSim processes at once has to give each of them a base index of its
+    own, or their calculation directories collide in the way this module's header describes. It
+    cannot simply number the processes, because there are usually far more setups to run than slots
+    to run them in, and a slot is reused as soon as the process occupying it finishes. What it needs
+    is a pool the size of the slot count, borrowed from for the length of one child and returned
```
k-rieck: some docstrings are for me a bit hard to read (doesn't sound like natural English to me sometimes). Something more understandable could be "What it needs is a pool with as many slots as the slot count. A slot is borrowed for the duration of a child and returned afterwards. That is what this class provides."

k-rieck (approved):

PR #628, opened 2026-09-03
Make the zenith clamp in the direct-normal irradiance actually apply (moves golden values)

noah80:
## Problem

`calculate_direct_normal_radiation` divides the direct horizontal irradiance by the cosine of the solar zenith angle and is meant to clamp the angle at `zenith_tol` (87??) first, because toward the horizon the cosine goes to zero and past it goes negative. The clamp was written as a chained assignment ??? `solar_pos["apparent_zenith"][mask] = tol` ??? which under pandas copy-on-write (always on in the pinned pandas 3) writes into a temporary and is discarded. pandas has been saying so in every golden log (`ChainedAssignmentError`, `weather.py:1400`). The clamp never applied: a small horizontal irradiance at 89?? became a very large normal one, and irradiance recorded past 90?? a **negative** one. Four weather readers call it, including DWD TRY ??? the default source of every golden setup.

## What was done

- One `.loc` assignment, which is the form that writes, with the reason in the comment.
- `zenith_tol` must lie strictly between 0 and 90; anything else is refused with a ValueError, since a clamp at or past 90 defeats the clamp and would come out as huge or negative irradiance rather than an error.
- The error for a NaN result names how many timesteps are NaN, the first of them, and whether the horizontal irradiance or the solar position was the NaN side, with the coordinates ??? instead of "Something went wrong...".
- Tests: a full day at Aachen with a constant non-zero horizontal irradiance; without the fix the night timesteps come out negative. It asserts no value is negative, none exceeds the clamped bound, every clamped timestep divides by the same cosine, and the highest-sun timestep divides by its own cosine (so a divisor returning a constant cannot pass). Two more tests cover the refused clamp angles and the NaN message.
- Also logged (not fixed): `PVSystemConfig.location` as F-3 in `roadmap/p4_random_findings.md` ??? never read by the component, no longer needed by the cache key once #625 lands (the key still carries it, being the whole config JSON), a P4-sweep item because it touches 26 call sites and 19 JSONs.

## Result ??? ?????? this PR moves reference numbers and needs a golden re-bless before it can merge

Base suite 4512 / 0 on the first two commits; the third (review fixes) changes no numbers ??? the Aachen annual DNI stays at 960.2 kWh/m?? (sanity threshold 950) ??? and passes test_weather, prospector, mypy. Golden: **all 16 pairs diverge, 42???68 KPIs each.** Annual PV production of the shared reference system 26 441.4 ??? 26 328.5 kWh (???0.43 %), the January week 83.3 ??? 82.5 kWh (???0.96 %; more in winter, as low sun should); self-sufficiency, storage and heat-distribution KPIs move by tenths of a percent; one peak KPI (heat-pump household, max grid draw, January week) moves 21.0 ??? 14.55 kW because a single dispatch instant shifted. The old references embed an overestimate of direct irradiance near the horizon. **To merge:** dispatch `golden-update.yml` on this branch and commit the blessed references ??? locally produced goldens are not canonical by the spec, and this is exactly the case that rule exists for.

The third commit answers the automated review of this PR. Its central claim ??? that the fix is a no-op and the test cannot catch a revert ??? came from a sandbox on pandas 2.3.3 with copy-on-write off; this repository pins pandas 3.0.5, where copy-on-write cannot be disabled, and there the old line yields 90 negative values over the test day and the new one none. Taken: the clamp-angle validation, the error message, the magnitude assertion, and two factual corrections to the F-3 note (seven references to the field, not three; #625 is open and the key still contains the field).


noah80:
Rebuilt on current main as a single commit. Main had renamed the function and its parameters to `calculate_direct_normal_irradiance_in_watt_per_square_meter` with `_in_degrees` suffixes, so the three original commits no longer applied; the same three changes and three tests are re-expressed on the new names. The findings note is now F-7 (F-3 is taken by the CHP summer branch) and its counts are re-verified against today's main. The revert check still holds on pandas 3.0.5: with the chained line back, the full-day test reports negative night-time DNI. Bless dispatched on this branch; the bless PR lands here, then this can merge.

PR #680, opened 2026-09-10
P4 gates D-2, D-8 and D-16: three retirements to obsolete/, and the physics decisions recorded

noah80:
﻿The first three gate decisions of the P4 component sweep, executed one per commit, plus the record of the six physics decisions taken in the same interview.

- D-2: the last zombie heat-pump controller (controller_l1_heatpump, zero users) moves to obsolete/.
- D-8: the fuel-cell controller, whose control methods read an undefaulted dataclass as class attributes and cannot run, moves to obsolete/ with its test and the three legacy configuration.py dataclasses it was the only user of, which unblocks the deletion of that legacy block.
- D-16, decided beyond the survey's recommendation: everything to obsolete/, nothing deleted. The MPC and PID controllers, the wind turbine and the price signal move with their tests; the storage controller and the four dead legacy configs are moved out of the files they shared into their own files; only the two dead factories are deleted. HouseholdWarmWaterDemandConfig stays, its class constants are live.
- The economic example's prose no longer describes the MPC controller as the library's price consumer.
- D-4, D-7, D-9, D-11, D-12 and D-21 are recorded as answered in the requirements document and the survey; their gate commits follow separately, each with its result diffs.

Verified: no setup instantiates any retired class, the energy-management system declares no default connection for them (the F-1 defect is not triggered), the full fleet re-records byte-identical, no golden or twin moves, and the quality gates already exclude obsolete/. The roadmap's two energy-system mockups still name the retired L1 controller, so the class-validation test now pins those entries as a missing module; repointing them is a batch-2 decision.


noah80:
Review round 1 addressed in b3af06fe; 2eb66c77 is a docs-only follow-up bringing §11 up to date with three later decisions.

- Files under `obsolete/` are frozen text by rule: the README now says each file is archived exactly as it stood on the day it was retired, import paths included, and is not importable in place. The moved controller and tests are deliberately left as they were (no rewiring, no commented-out imports).
- The MPC config figure was wrong and is corrected in all six places: 27 of its 49 fields are lists (13 result buffers, 12 forecast inputs, 2 coefficient vectors), not 31 runtime buffers.
- The fuel-cell configuration docstring now names the retired test as the other user of the three classes.
- The deletion of `RandomNumbersConfig.get_default_config` is kept and its consequence recorded: the JSON default-construction path for RandomNumbers is closed until the conversion mints its preset; both in-repo scenarios pass explicit blocks.
- casadi is retired with the MPC controller: removed from requirements.txt, mypy.ini and mypy_moderate.ini. mypy is clean in both profiles.
- Declined: deprecation shims or re-exports for the moved modules and classes (the obsolete/README table is the record, per the intended-break policy); restoring the deleted factories; deleting the retired tests. The orphaned 5R1C singleton writes in the Building are D-32 part two, the next PR after this one.

PR #681, opened 2026-09-10
The parity summary keeps the checker's exit code without a pipeline, because the container runs sh

noah80:
Run 34504646328 passed all 22 parity triples and still went red: the summary step runs inside the test container, where GitHub uses sh, and sh rejects the bash-only set -o pipefail before the table is printed. The step now takes the freshness job's form: the checker writes to a file, its exit code is kept, the file is shown and appended to the step summary. No other workflow carries the pattern. Verified with an empty verdicts directory that the redirect form still exits 1 when the checker does.

### DIFF
--- a/.github/workflows/p3-parity.yml
+++ b/.github/workflows/p3-parity.yml
@@ -169,20 +169,15 @@ jobs:
           path: verdicts
           merge-multiple: false
       - name: Print one table covering every triple
-        # No pipeline here: GitHub runs this step with 'sh' inside the test container, which
-        # rejects the bash-only `set -o pipefail` ("Illegal option -o pipefail") and expands
-        # PIPESTATUS to "Bad substitution". Redirect, then show the file, so the step keeps the
-        # checker's own exit code — the summarize path returns 1 when any triple missed parity
-        # or when no verdicts were collected at all, and `tee` would otherwise replace that
-        # with its own 0.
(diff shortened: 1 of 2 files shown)


noah80:
Review round 1 addressed in 34b91ee3.

The review established that the test container has bash: the resource-monitor composite steps of the same job run under `shell: bash` and ran before the summary step failed. The plain `run:` step fell back to sh, which is an observation about this container, not the rule the earlier comment claimed.

- The step now declares `shell: bash` and keeps the original two lines, `set -o pipefail` and the checker piped into `tee`. The seven-line sh idiom goes, and with it the two defects found in it: `set -e` re-enabled before the `cat` calls (a failing cat would have replaced the checker's exit code) and `2>&1` folding tracebacks into the rendered summary.
- The comment states the observed fallback and why pipefail is needed.
- New test `test_summarizing_a_dispatch_exits_nonzero_unless_every_collected_triple_reached_parity` pins the `--summarize` exit codes (0 at full parity, 1 when a triple missed parity, 1 when nothing was collected), writing the verdict files through `TripleVerdict.to_json()` so a change to the verdict shape reaches the test.
- Declined: making the checker write the step summary itself and extracting the sh idiom into a composite action, both moot under `shell: bash`. The two freshness workflows carry the same idiom and get the same treatment in a separate follow-up PR.

PR #682, opened 2026-09-10
A bless keeps every value the gate would have accepted, so its diff says only what moved

noah80:
﻿The first three gate decisions of the P4 component sweep, executed one per commit, plus the record of the six physics decisions taken in the same interview.

- D-2: the last zombie heat-pump controller (controller_l1_heatpump, zero users) moves to obsolete/.
- D-8: the fuel-cell controller, whose control methods read an undefaulted dataclass as class attributes and cannot run, moves to obsolete/ with its test and the three legacy configuration.py dataclasses it was the only user of, which unblocks the deletion of that legacy block.
- D-16, decided beyond the survey's recommendation: everything to obsolete/, nothing deleted. The MPC and PID controllers, the wind turbine and the price signal move with their tests; the storage controller and the four dead legacy configs are moved out of the files they shared into their own files; only the two dead factories are deleted. HouseholdWarmWaterDemandConfig stays, its class constants are live.
- The economic example's prose no longer describes the MPC controller as the library's price consumer.
- D-4, D-7, D-9, D-11, D-12 and D-21 are recorded as answered in the requirements document and the survey; their gate commits follow separately, each with its result diffs.

Verified: no setup instantiates any retired class, the energy-management system declares no default connection for them (the F-1 defect is not triggered), the full fleet re-records byte-identical, no golden or twin moves, and the quality gates already exclude obsolete/. The roadmap's two energy-system mockups still name the retired L1 controller, so the class-validation test now pins those entries as a missing module; repointing them is a batch-2 decision.

### DIFF
--- a/.github/workflows/golden-update.yml
+++ b/.github/workflows/golden-update.yml
@@ -6,8 +6,10 @@
 # The bless is sticky: a KPI whose fresh value the gate itself would still accept
 # keeps the value already committed, and a golden that comes out identical is not
 # rewritten, so the PR diff carries only what actually moved. When nothing moved at
-# all, the diff is the manifest alone. Someone who deliberately wants the stored
-# last-digit noise replaced by fresh values runs the script with --force-rewrite.
+# all, the diff is the manifest alone. Nothing is ever dropped: a KPI the run no
+# longer produces stays in its golden and the gate goes on reporting it as missing.
+# The force_rewrite input dumps the fresh values verbatim instead — the way to clear
+# accumulated noise, to retire a KPI, and to repair an unreadable golden.
 
 name: golden-update
 
@@ -18,6 +20,11 @@ on:
         description: "Why are you re-blessing the goldens?"
         required: true
         default: "intentional simulation-output change"
+      force_rewrite:
+        description: "Overwrite every stored value with the fresh one, dropping KPIs the runs no longer produce."
+        type: boolean
+        required: false
+        default: false
 
 permissions:
   contents: write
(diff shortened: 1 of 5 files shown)


noah80:
Review round 1 addressed in fa05c516. This PR is an enhancement of the bless, not a bug fix; the sticky keep is new behaviour and the tests pin it.

- An unusable stored golden (unreadable, not an object, or not flat) now fails that pair like a run error: named in the log, counted as errored, non-zero exit. Only an absent file means "write verbatim". `--force-rewrite` bypasses the load and is the sanctioned repair path.
- Keys the fresh run no longer produces are never dropped: they stay in the golden with their stored value and are always named as absent, so the gate keeps reporting them as missing until a deliberate forced bless retires them.
- `--force-rewrite` is reachable from CI through a new `force_rewrite` dispatch input; the bless PR body says when a bless was forced.
- `merge_into_golden` returns a structured record (moved, new, absent, kept); one formatter prints the line, so the counts are exact (moved no longer includes new or absent keys) and the tests assert the record and the file content instead of prose. `write_golden` decides "rewritten" on the parsed mapping, so a value-identical but differently formatted file is left alone and read once.
- Wording: the file always carries the full mapping and only its diff is limited; the keep threshold is the gate's default tolerance; the workflow says a manifest-only diff means "reproduced within the gate's tolerance"; the local-inspection note says a local run is sticky by default.
- Declined: a shared tolerance predicate with golden_check (CI runs the gates at defaults, and the bless serves the same goldens), nondeterministic-pair handling (no configuration sets the flag), a return-type shim for `write_golden` (no caller outside the script), and removing `--force-rewrite` (the opposite finding, that it was unreachable, was the real one).

PR #683, opened 2026-09-10
The CHP controller's buffer factories share one helper; the thresholds stay as they were

noah80:
﻿## D-4: the CHP controller's buffer factories share one helper; the thresholds stay as they were

The four default configurations of `L1CHPControllerConfig` cross a fuel with a buffer storage. The two buffer factories now derive from their fuel's base factory through one pure helper that applies the shared buffer parts and takes the per-fuel lower bounds. The original 2023 thresholds are kept unchanged and their asymmetry is recorded as unexplained. A `__post_init__` check refuses a band whose lower bound is not strictly below its upper one, and the four factories are pinned as literals in the tests. No setup, energy-system file or golden reference instantiates the class.

### DIFF
--- a/hisim/components/controller_l1_chp.py
+++ b/hisim/components/controller_l1_chp.py
@@ -30,11 +30,39 @@ __maintainer__ = "Vitor Hugo Bellotto Zago"
 __email__ = "vitor.zago@rwth-aachen.de"
 __status__ = "development"
 
+#: Lower set temperature of the buffer storage, given in °C.
+BUFFER_T_MIN_HEATING_IN_CELSIUS = 35.0
+#: Upper set temperature of the buffer storage, given in °C.
+BUFFER_T_MAX_HEATING_IN_CELSIUS = 40.0
+#: Julian day the heating season begins on when a buffer storage is present.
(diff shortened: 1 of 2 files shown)


noah80:
﻿Review round 1 addressed in 5726904e.

- The normalisation is reverted: all four factories keep their 2023 thresholds (chp 42/60, fuel cell 50/60, chp-with-buffer 35/40 and 50/60, fuel-cell-with-buffer 31/40 and 42/60). The asymmetry is recorded in the class docstring as unexplained and deliberately kept rather than guessed at.
- The buffer helper stays, now pure (`dataclasses.replace`) and parameterised by the per-fuel lower bounds; the shared parts (upper bound 40, season begin one day early) are module-private constants, with the season begin derived from one base constant.
- `fuel_cell_with_buffer` keeps the default name FuelCellController; all four default names are pinned in a test.
- `__post_init__` refuses a band whose lower bound is not strictly below its upper one.
- The hydrogen-threshold docstring is corrected; the four factories are pinned as explicit literals instead of the cross-factory comparison.
- Not changed: merging the two base factories (the conversion batch replaces them with presets) and the summer-branch comparison against the heating band's upper bound, which is filed as a separate defect.

PR #684, opened 2026-09-10
D-12: a scaled PV system records the share it was scaled by, so its record re-executes

noah80:
﻿P4 decision D-12, option (a): a scaled PV system records the share it was scaled by, so a realized record re-executes to the same power. The physics fix turned out to be on main already: #638 stamped the real share onto the scaled config because its half-PV probe column needed it, and the survey's entry had gone stale. This commit finishes the decision around that fix: a round-trip test through the record's own writer and codec (a half-share array comes back identical), a docstring stating that the default factory's power argument is the array's maximum and a recorded pair must never be fed back through it, the survey corrected to record (a) as executed, and one sentence in the RenoVisor spec. No number moves: every caller of the scaled factory forwards a share of 1.0, RenoVisor can only express 0 or 1, and the full fleet re-records byte-identical.

### DIFF
--- a/hisim/components/generic_pv_system.py
+++ b/hisim/components/generic_pv_system.py
@@ -16,7 +16,7 @@ economic and CO2-footprint figures used in post-processing.
 import datetime
 import enum
 import os
-from dataclasses import dataclass, field
+from dataclasses import dataclass, field, replace
 from typing import Any, Dict, List, Optional, Tuple
 
 import numpy as np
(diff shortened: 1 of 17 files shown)


noah80:
Review round 1 addressed in 69408eb2. As the review noted and the commit body says, the recording fix itself landed on main in #638; this PR contributes the proof, the contract and the corrections below.

- `share_of_maximum_pv_potential` is range-checked in `__post_init__` (a share outside [0, 1] is refused naming the value), and `get_scaled_pv_system` stamps the share through `dataclasses.replace` so the check runs on that path too.
- The factory argument that names the array's maximum is now `maximum_power_in_watt`; the `power_in_watt` field stays the scaled result. Thirteen call sites renamed; no recorded twin or golden moves (`scripts/record_all_setups.py --check` confirms).
- Docs: the RenoVisor spec says no simulation result changes while the recorded provenance for a zero-kWp payload goes from 1.0 to 0.0; the survey disclaimer is narrowed to what was re-verified, and the PV entry records the rename so the conversion's preset and law do not reintroduce a maximum named `power_in_watt`.
- The three long docstrings are trimmed to contract and reason; the record round-trip sequence shared with the occupancy test becomes one helper in `tests/functions_for_testing.py`.
- Declined: renaming the field itself, a guard against replaying a recorded pair, retitling the PR, and driving the round-trip through a full re-run.

PR #685, opened 2026-09-11
A constructor called from a file gets its arguments decoded

noah80:
﻿## D-19: a constructor called from a file gets its arguments decoded

A named constructor under `constructor:` had its YAML arguments forwarded verbatim, so `location: AACHEN` reached `WeatherConfig.for_location` as a string and the author was told `'str' object has no attribute 'value'`. Only the all-scalar TABULA constructor worked from a file.

- `codec.py`: the annotation-driven decoding becomes `decode_argument`, shared by `config` blocks and constructor calls, with a list case for `List[X]` / `Union[X, List[X]]`.
- `configure.py`: `_call_builder` decodes each argument against the parameter's resolved annotation; a bad value is EF-1A at `components.<name>.constructor.<constructor>.<parameter>` with the members and a did-you-mean.
- `presets.py`: `@constructor` refuses at decoration time a parameter whose type no written value can become (allowed: scalars, enums, dataclasses with `from_dict`, and Optional/Union/List over those).
- New `tests/test_energy_system_constructor_arguments.py` calls all four existing constructors from YAML.

Golden-neutral: no recorded twin uses `constructor:`.

### DIFF
--- a/energy_systems/README.md
+++ b/energy_systems/README.md
@@ -66,7 +66,7 @@ provides, every fact somebody reads, and which provider each read resolved to.
 
 A constructor's arguments are written and decoded exactly like a `config` value: an enum by its
 member name or its value, a nested object as a mapping of its own fields, several of them as a
-list, and a wrong one refused by name with the spellings that would have worked.
+list, and a wrong one refused by name, listing the accepted members or types.
 
 Editors get the same knowledge from `hisim/energy_system_v3.schema.json`, which every file in this
 directory binds to with its first line. The schema is generated from the same declarations
--- a/energy_systems/README.md
+++ b/energy_systems/README.md
@@ -64,6 +64,10 @@ file and without running it, the file's knobs — a flag per group and a selecte
 which is everything a consumer of a checked-in system edits — and then every fact somebody
 provides, every fact somebody reads, and which provider each read resolved to.
 
+A constructor's arguments are written and decoded exactly like a `config` value: an enum by its
+member name or its value, a nested object as a mapping of its own fields, several of them as a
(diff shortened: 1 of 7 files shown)


noah80:
Review round 1 addressed in 901708e0.

- One rule for what a written value can decode into: `is_decodable_annotation` in `hisim/config/presets.py` is now consulted by the codec as well as by the `@constructor` gate. It admits None, the four scalars, any Enum, any dataclass (with or without `from_dict`, matching the codec's fallback), `List[X]`, and unions with at most one dataclass member and one list member. A shape-matrix test pins gate and decoder to the same answer for every shape.
- Two decoder bugs fixed: `null` for an `Optional[Enum]` parameter builds instead of being refused, and `Union[str, Enum]` accepts a free string where `str` is admitted.
- Constructor arguments are checked for shape after decoding (null for a required parameter, a list where a single value belongs and vice versa, a scalar where an object belongs) and refused at the argument's own key path. The sparse-override path keeps its pass-through, since the class's `from_dict` follows there; a complete `config` block never went through this path, and the module header now says so.
- A constructor parameter whose annotation cannot be resolved is refused at class creation, naming the parameter and the annotation; the inert sentinel is gone. Resolved parameter types are computed once at decoration and carried on the builder; `configure` and `schema_export` both read them, removing the duplicate resolver. The regenerated schema is byte-identical.
- Tests: the by-value enum test asserts identity (the enums are str subclasses, so `==` masked a non-decode); hand-derived assertions join the constructor oracle; a field-path test pins that a `List[float]` override now decodes item by item, which used to be refused.
- Declined: dropping the "Did you mean" assertions (the suggestion is part of the message contract) and the red-before-green remark.

PR #686, opened 2026-09-11
The report region comes from the Weather's config, not from a global

noah80:
﻿## D-32 (part one): the report region comes from the Weather's config, not from a global

The Weather wrote its location into the process-wide singleton under `SingletonDictKeyEnum.LOCATION` at construction time and postprocessing read it back as the report region in two places. Across two simulations in one process that key was not necessarily this run's.

- `postprocessing_main.py`: one helper `region_of(ppdt)` finds the Weather among the run's wrapped components and returns its configured location; both readers (pyam export, config JSON) call it. No Weather yields `""`; two Weathers report the first.
- `weather.py` stops writing the key; `SingletonDictKeyEnum.LOCATION` is deleted.
- New `tests/test_postprocessing_region.py`; `test_singleton_sim_repository` adjusted.

Report metadata only: no golden carries a region field. The six 5R1C Building keys (part two) wait for the D-16 gate branch.

### DIFF
--- a/hisim/components/weather.py
+++ b/hisim/components/weather.py
@@ -656,7 +656,6 @@ class Weather(Component):
             my_display_config = DisplayConfig()
         self.last_timestep_with_update = -1
         self.weather_config = config
-        SingletonSimRepository().set_entry(key=SingletonDictKeyEnum.LOCATION, entry=self.weather_config.location)
         self.parameter_string = my_simulation_parameters.get_unique_key()
 
         self.my_simulation_parameters = my_simulation_parameters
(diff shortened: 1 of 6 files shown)


noah80:
Review round 1 addressed in fb97de5c.

- A run with several Weathers is reported under all of their locations, joined in component order with " / ", so the region is deterministic and no station is silently dropped; no Weather still gives an empty region, one Weather its location. Today every recorded system has exactly one Weather.
- The docstring is trimmed to that contract and names the second writer correctly: the scenario-evaluation config JSON, not the webtool result JSON.
- Tests assert literal station names (Aachen, Madrid, and both join orders) rather than the object the helper reads from; the private data-transfer helper is replaced by the postprocessing test framework's own, which gained an optional component list; the assertion that the enum member is absent is dropped as structural.
- Declined: the finding that the `location: str` annotation is redundant (without it, mypy reports no-any-return under this repository's configuration; verified), a deprecation shim for the removed enum member, a warning for runs without a Weather (the simple setups legitimately have none), and a consumer-level pyam assertion. The adjacent scenario-name and description reads stay on the singleton: they are written by the runner, not by a component at construction time, and are outside D-32.

PR #687, opened 2026-09-11
Operation modes become enums, so a typo cannot silently change a controller

noah80:
﻿## D-27: operation modes become enums, so a typo cannot idle a controller

Three hydrogen-chain controllers carried `operation_mode: str` over a closed set of strings, branched with if/elif and a silent fallback, so a misspelled mode produced a plausible-looking run.

- `PtxOperationMode` (4 members), `XtpOperationMode` (2) and `RsocBatteryOperationMode` (3), one per module because the three vocabularies differ; declared as `(str, Enum)` like `BoilerType`.
- Member values are the old strings, so any stored configuration still decodes; the wire format is unchanged.
- The old catch-all fallbacks, reachable only through a typo, now raise `ValueError` naming the mode.
- Tests switched to members; six new round-trip tests pin the wire value.

Golden-neutral: the one live hydrogen setup uses none of the three controllers. The RSOC controller is scheduled for `obsolete/` under D-25 and takes its enum with it.

### DIFF
--- a/hisim/components/controller_l2_ptx_energy_management_system.py
+++ b/hisim/components/controller_l2_ptx_energy_management_system.py
@@ -64,6 +64,28 @@ class PTXControllerConfig(ConfigBase):
     standby_load: float
     operation_mode: PtxOperationMode
 
+    def __post_init__(self) -> None:
+        """Normalises the operation mode into a :class:`PtxOperationMode` member.
+
+        The mode is wire format: a configuration read from JSON, from HDF5 or written by
+        hand arrives carrying the plain string the field has always been serialized as,
+        while a caller in Python passes the member. Both are accepted here and both leave
+        as the member, so only one kind of value ever reaches the control law. A value
+        that names no mode is refused where it was written, instead of travelling into a
+        controller that has no branch for it.
+
+        Raises:
+            ValueError: For an ``operation_mode`` that is neither a member of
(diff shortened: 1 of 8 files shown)


noah80:
Review round 1 addressed in a6170bfe.

- Report rendering fixed once for every config: `ConfigBase.get_string_dict` renders enum values by their wire value (also inside lists, dicts and tuples), so the report prints `Operation mode: NominalLoad` and, repository-wide, `Boiler type: CONDENSING` instead of the qualified member name. No golden or recorded twin carries report text; no existing test had pinned the old spelling.
- Origin-level validation: the three configs normalise `operation_mode` in `__post_init__` (a valid string becomes the member, an unknown value is refused naming every mode), so a typo fails where it is written rather than at the first timestep; `from_dict` and `dataclasses.replace` both run it, pinned by tests.
- Tests: the refusal tests moved from injecting a bare string into `system_operation` to constructing a config with a wrong value; the round-trip tests iterate the literal legacy strings instead of feeding the enum back to itself. The unconditional else in `system_operation` stays as the guard for a future member.
- Wording: the old catch-all never idled a controller, it silently followed the load; the docstrings now say a typo can no longer silently change what the controller does.
- Declined: making the enums private or test-only (they are the field types and therefore wire vocabulary), merging the three (per-module enums is the D-27 decision), renaming members to match values, the per-timestep cost note, and new arithmetic tests for the nominal and minimum modes.

PR #688, opened 2026-09-11
The example's magic literal becomes a law, and the template shows one

noah80:
﻿## D-31: the example's magic literal becomes a law, and the template shows one

`example_template.py` is the file every new component author copies, and it demonstrated none of the sizing mechanism. `example_component.py` hid a law in a literal: `capacity = 45 * 121.2` is 45 J/K per m² times the conditioned floor area of the default TABULA building.

- `example_component`: `capacity` is a `sized_field` over `Size.CONDITIONED_FLOOR_AREA_IN_M2`; the factory says `AUTO`; the duplicate literal in `build()` is gone. Resolved against the default building the value is 5454.0, exactly as before, asserted in the test.
- `example_template`: one sized field with its law and note, an `AUTO` factory, a numbered walk-through of the five steps, and a comment block with the exact `SIZING_CONTRIBUTIONS` shape pointing at the real providers. No contribution entry: the kernel only accepts facts that are `SizingContext` fields, and a template-only fact in the shared vocabulary was rejected in favour of the documented block.
- Tests resolve against a stated `SizingContext(conditioned_floor_area_in_m2=121.2)` through a shared helper; `describe` shows the sizable field with its law text.

Golden-neutral: neither class is in any setup or energy system. Schema unchanged.

### DIFF
--- a/hisim/components/example_component.py
+++ b/hisim/components/example_component.py
@@ -10,7 +10,7 @@ from dataclasses_json import dataclass_json
 # Owned
 from hisim.simulationparameters import SimulationParameters
 from hisim.component import Component, SingleTimeStepValues, ComponentInput, ComponentOutput
-from hisim.config import ConfigBase, ComponentID, DisplayConfig
+from hisim.config import AUTO, ConfigBase, ComponentID, DisplayConfig, Sizable, Size, concrete, sized_field
 from hisim import loadtypes as lt
 from hisim.economics.facts import CostRelevance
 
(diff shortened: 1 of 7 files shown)


noah80:
Review round 1 addressed in ec052639.

- The template's rated-power cap stays and is now honest about units: the capped output channel is declared in watts (it was declared in watt-hours while capping watts by watts), and two tests pin it, one where a small floor area makes the cap bind and one where it does not.
- Both sized fields carry `value_type=float`, so a string written in a file is coerced to a float or refused at decode time instead of reaching arithmetic; both behaviours are pinned.
- The field notes are built from the module constants rather than repeating them, so editing a constant cannot leave its note stale.
- Docstrings corrected: the note is printed by `describe`, not by the audit trail; the pointed-at components demonstrate a sibling scale and a rounding, and no component uses `.at_least`, so that example points at the sizing tests; step 4 of the walkthrough says where a contribution is declared rather than claiming the template declares one.
- Tests: a base-marked assertion pins the resolved default capacity (it was only checked in a slow suite); the remaining literal floor areas use the named test constant; the describe tests read the structured field description instead of the rendered law string, keeping one rendered assertion for the template itself since what `describe` prints is that file's purpose.
- Declined: restoring the optional-with-literal default and field order (the accepted cost of D-31), a positivity guard on the floor-area law, deriving the default area from the TABULA lookup in the fast tests, trimming the tutorial docstring, and a real contribution entry.

PR #694, opened 2026-09-11
D-30: the smart device that cannot be built moves to obsolete/

noah80:
## D-30: the smart device that cannot be built moves to obsolete/

`hisim/components/generic_smart_device.py` raises `KeyError: 'utsp_reports'` on every construction: its `build()` reads a `HISIMPATH` key that no longer exists. It has no setup and no test, and the preset the naming supplement proposed for it would have shipped a placeholder identifier matching no device. Under the D-16 rule it is moved, not deleted.

- The module moves to `obsolete/components/` unchanged (git rename, no content change).
- Its registry rows in the economics adapter and the extraction test go, as in #680; the docs module list drops the entry.
- Left in place, as vocabulary: `ComponentType.SMART_DEVICE` (named by the energy management system's consumer list), the KPI tag, and the `SMARTDEVICESINCLUDED` singleton key, which had neither writer nor reader already and is D-32's business.
- Records: the obsolete README row, the §11 execution note, the R3 row, and the `smart_devices_included` copy-law candidate struck in the sizing-fact inventory.

No setup, recorded twin or golden reference names the class.


noah80:
Review round 1 addressed in 78f7cffc.

- The R3 legend now says what `del` has meant since D-16: retired, moved to `obsolete/`, never deleted. The rows keep their token.
- Declined: a regression test for a move; "fixing" the archived module's KeyError (files under `obsolete/` are archived as they were); shims or removal for the `ComponentType`, KPI tag and singleton-key vocabulary that named the component (kept as vocabulary, as #680 kept the wind turbine's); the decision-versus-omission reading of the removed `BY_CLASS_NAME` row (answered in #680); and the survey passages that still describe the pre-move path, which are dated history.

PR #695, opened 2026-09-11
The Building stops writing six thermal-model values nobody reads

noah80:
## D-32, part two: the Building stops writing six thermal-model values nobody reads

The Building wrote six 5R1C values (glazing, surface-to-air, opaque, ventilation transmission coefficients and the envelope capacity) into the process-wide singleton at build time. Their only readers were the PID and MPC controllers, which #680 moved to `obsolete/`; a tree-wide grep finds no live reader left.

- The six `set_entry` writes in `building.py` and the six enum members go; a comment holds the numbers 8 to 13, as #686 did for the LOCATION key, so nothing is renumbered.
- The comment describing the feed to the two controllers is corrected; the singleton test's presence-and-value check is repointed at the Weather's forecast series that a run still writes.
- Left in place: the three heat-flux forecast keys the Building writes only under `predictive`. They are reader-less too, but they are the tail of the whole-year forecast computation; removing them would retire that path, which this decision does not cover.

Result-neutral: the values were only written to the singleton and never reach a result series, so no golden or twin moves. Part one (the LOCATION key) is #686.

### DIFF
--- a/hisim/components/building/building.py
+++ b/hisim/components/building/building.py
@@ -913,33 +913,6 @@ class Building(cp.Component):
             self.thermal_conductance_by_ventilation_in_watt_per_kelvin,
         ) = self.get_conductances()
 
-        # send building parameters 5r1c to PID controller and to the MPC controller to generate an equivalent state space model
-        # state space represntation is used for tuning of the pid and as a prediction model in the model predictive controller
-        SingletonSimRepository().set_entry(
-            key=SingletonDictKeyEnum.THERMALTRANSMISSIONCOEFFICIENTGLAZING,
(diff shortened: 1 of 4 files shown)


noah80:
Review round 1 addressed in 2701d162.

- The singleton test asserts each surviving key once: the redundant membership checks are gone, the non-emptiness checks on the Weather's two yearly forecast series stay, and the docstring says what is proven (a run populates the singleton with series the Weather writes unconditionally).
- Declined: the six deleted enum members being referenced by files under `obsolete/` (archived as they were, not importable in place, per the obsolete README); a deprecation shim for the removed members (intended-break policy, recorded in §11); renumbering the surviving members (the gap is commented on purpose); the survey passage that still describes the keys as present (dated history).

PR #696, opened 2026-09-11
D-29: the plainer electrolyzer pair and its controller move to obsolete/

noah80:
## D-29: the plainer electrolyzer pair and its controller move to obsolete/

HiSim had two electrolyzer classes and two hydrogen-storage classes for the same two devices, neither pair used by a setup. The owner keeps the `_and_h2_storage` pair (`AdvancedElectrolyzer`, `HydrogenStorage`: the richer model with waste energy and part-load) and retires the `generic_*` pair. Under the D-16 rule everything is moved, not deleted.

- Moved to `obsolete/components/` unchanged: `generic_electrolyzer.py`, `generic_hydrogen_storage.py` and `controller_l1_electrolyzer.py` (written only for the retiring electrolyzer); their tests `test_generic_electrolyzer.py` and `test_h2storage.py` move to `obsolete/tests/`. `test_controller_l1_generic_electrolyzer.py` stays: despite its name it tests the surviving hydrogen controller.
- `AdvElectrolyzerConfig` and `HydrogenStorageConfig`, both without callers, move out of `configuration.py` into `obsolete/components/configuration_hydrogen.py`; `configuration.py` keeps everything else.
- The CHP controller drops its optional default connection from the retired storage rather than getting a new one; its optional hydrogen-SOC input stays.
- Registry rows in the economics adapter and the extraction test, and the docs module entries, follow as in #680. D-28 (a constructor for the retiring config) is confirmed moot.

Result-neutral: no setup, twin or golden names any moved class; the live electrolyzer setup uses the third, manufacturer-table electrolyzer, which stays.


noah80:
Review round 1 addressed in a76ec0fe.

- The obsolete README now counts three electrolyzer models and names the live third, `generic_electrolyzer_h2.Electrolyzer`; the §11 execution note says "no module or class was deleted", since a default connection, two registry rows and three docs entries did go, as the same note states.
- Declined: the moved modules importing each other and their test by their old paths (archived as they were, not importable in place); the `# ToDo` marker carried along in the archived config file; the archived configs "duplicating" the survivors (they are the dead copies the decision retired); the retired tests leaving the suite; and the CHP controller keeping its optional hydrogen-SOC input without a default source, which is the owner's D-29 follow-up decision (the input is guarded and skipped when unconnected).

PR #699, opened 2026-09-11
D-7: four square metres of solar collector per apartment

noah80:
## D-7: four square metres of solar collector per apartment

`SolarThermalSystemConfig.area_m2` becomes a sizing law, `4 m² × number_of_apartments`, so the three solar-thermal setups stop doing that arithmetic by hand (one of them passed 4 unmultiplied). An explicit area still overrides the law.

- The law reads the setup's dwelling count, the same number the setups scale their DHW storage by; for the default single-family building it yields the 4 m² the setups used, so no golden moves and no bless is needed.
- With the law live, the gas solar-thermal setup's two-dwelling probe records an 8 m² collector, so the collector joins the DHW storage as an override group in that setup's grouping; the grouping overview and the three scenario JSONs are regenerated by their tools.
- The manually-calculated-capex factory keeps a plain area default: its CO2 derivation is affine and the law algebra has no addition, and the factory has no caller.
- Tests: the law resolves to 4 m² for one apartment and 12 for three; an explicit area wins.

Twins re-recorded; the recorder confirms they match. `describe` shows the field with its law and note.


noah80:
Review round 1 addressed in bbcb2dc2.

- The dead `get_default_solar_thermal_system_manually_calculated_capex` factory and its CO2 helper are deleted (zero callers; a dead factory is deleted under the D-16 rule), which removes the diverging default and the AUTO hazard in its capex arithmetic. One factory remains.
- The now-sizable area is read through `concrete()` everywhere: the component reads it once at construction and uses that in `i_simulate`; the economics adapter does the same. A new test asserts a three-dwelling collector produces three times the single-family power for one timestep.
- The constant's comment states the law's assumption (demand scales with dwellings) as an assumption; the §11 answer paragraph carries a dated correction that no golden moved, and the executed note names the one remaining factory; the adapter contract scan references the constant instead of a literal.
- On "the law never fires on the declarative path": a recorded twin is a realized record and writes every sized value concretely, the PV power and heating load included, so it re-executes without consulting a law. The law fires for a hand-written file that omits the field and for every Python setup; the two-dwelling probe in the grouped twin carries the 8 m² it produced.
- Declined: a positivity guard on the apartment count (the Building never contributes one); reverting the AUTO default or the law to a plain float (the D-7 decision); making the constant private; a setup-level scaling test (the grouping probe test already pins the 8 m² record); the pass-through test (it pins that an explicit value beats the law).

PR #700, opened 2026-09-11
D-9: the buffer storage is sized from the generator, not the load

noah80:
## D-9 (C11): the buffer storage is sized from the generator, not the load

The buffer storage's volume is litres per kilowatt of the heat generator (20, 40 or 50 by generator kind), but all twelve setups passed the building's heating load into the parameter named for the generator's power. Each setup now passes its generator's maximal thermal power (the resolved boiler's, or the heat pump's rated output). No preset, no new field; physics only.

- Twins re-recorded: the gas, gas-solar-thermal and hydrogen sizers go from 155.62 to 171.18 l (+10 %, their boiler carries the 1.1 DHW uplift); `basic_household_only_heating` goes to 240 l (+54 %, its boiler is a fixed 12 kW preset). The four heat-pump setups are unchanged because their heat pump is sized at exactly the load.
- Measured week deviations before the bless: about 40 KPIs per changed setup, dominated by storage standby losses and investment; fuel consumption moves by less than 0.2 %. All changed setups are in the golden gate; the goldens are re-blessed by the golden-update workflow on this branch.
- A test pins the arithmetic (8558.8 W at 20 l/kW is 171.18 l, against the old 155.62 l).

Finding recorded for a separate decision: the oil, pellets and wood-chips sizers and `household_gas_solar_thermal` resolve their boiler without `number_of_apartments` in the context, so the DHW uplift is skipped although their controllers prepare hot water; that is why only three of the five boiler sizers the survey predicted have moved.


noah80:
Review round 1 addressed in b61930d5.

- The factory docstring states the contract in five lines instead of narrating the defect: the argument is the heat generator's maximal thermal power (a boiler's `maximal_thermal_power_in_watt`, a heat pump's `set_thermal_output_power_in_watt`), the litres-per-kilowatt figures are per kilowatt of installed generator power, and where a setup sizes its generator at exactly the load the two coincide. The "never the building's heating load" claim is gone: it was false for the four heat-pump setups by design and for four boiler setups by a defect.
- That defect is F-6, found by this PR's own execution and confirmed by three reviewers: the oil, pellets and wood-chips sizers and `household_gas_solar_thermal` resolve their boiler without the apartment count, so the DHW uplift is skipped and their generator equals the load. The owner decided to fix those four contexts as a separate PR with its own bless, after this one merges.
- The stale "60 l per kW" comment beside the 20 l/kW branch is corrected; the R3 row's twin count is reconciled (3 moved by 10 %, 1 by 54 %, 8 byte-identical); the arithmetic test states its inputs and outputs as literals instead of deriving the input from the boiler's law, which `tests/test_sizing.py` already pins.
- Declined: renaming the parameter (it names the generator); a grouped-twin test (the grouping test realizes every grouped twin against its flat twin and is what caught the stale record); the heat-pump path being a no-op (its heat pump is sized at exactly the load; stated in the PR body); the `concrete()`-returns-`Any` observation (kernel-wide, not this PR); the performance note (the larger volume is the intended result).

PR #716, opened 2026-09-12
The energy manager creates a target output only for a device it has, and names it by what it is

noah80:
F-1 (
oadmap/p4_random_findings.md): the energy manager created a target output for every device it could ever meet, inside the methods that merely describe its default connections, and named each port by its position in the output list. A district-heated house carried five ports for a heat pump, an electric heater and a solar collector it does not have, and retiring one dead method renamed every port after it fleet-wide.

- Describing a connection creates nothing. A DynamicComponentConnection may carry a DynamicComponentTargetOutput; the legacy auto-wiring grows that port beside the input, once per source device the setup built. The declarative path already worked this way and is unchanged.
- A dispatch port is named prefix + source weight (LoadingPowerInputForBattery_6); a duplicate raises. Legacy names stay distinct from the declarative DispatchTo… names, so the P3 parity table keeps its job; its EMS rows are respelled.
- Ports grown while wiring are registered by the wrapper before the values vector is sized (
egister_outputs_grown_while_wiring). EMS dispatch columns therefore sit at the end of the result column order; values are unchanged.
- json_generator.py asks the declarations which targets are default-grown instead of parsing OutputN.

District heating sizer: 14 → 9 declared EMS ports; heat pump sizer: 14 → 11. Registered result columns unchanged. All 12 flat and 11 grouped EMS twins re-record identically; week goldens for household_heatpump_building_sizer and dynamic_components pass. No bless needed. Scenario JSONs untouched (retired by #708; the freshness job is red until rebase).

Checks: EMS, dynamic-component, parity, dispatch and setup tests (92 passed, 3 new); flake8, prospector, mypy.


noah80:
Review follow-up, commit 6505fbdb. Of the 26 deduplicated findings, 17 were taken:

- The scenario JSON writer no longer rebuilds port names by string arithmetic: a grown port's bookkeeping entry records its prefix and whether a default connection grew it; the writer reads that, and raises by name on a port that was not built from a prefix and a weight (the declarative `DispatchFor…` ports).
- A feed the connection allows to stay unconnected grows nothing when the device does not publish the measured field, so a DHW-less heat pump gets no dead DHW target. The whole feed is skipped rather than the target alone, because the manager pairs inputs and targets one to one by weight.
- Duplicates are refused by meaning: one shared helper, used by both the imperative and the declarative creation path, rejects a second port with the same component type and weight, and the name guard now names all three causes, including two instances of one class through default connections, with the two supported routes (hand wiring with distinct weights, or the declarative format). That case was never supported: the legacy default connections give a class one fixed weight, so two instances shared a phantom and an ambiguous dispatch.
- A target whose tags do not carry the feed's component type is refused at growth.
- Four new tests: a port grown while wiring becomes a result column through `prepare_calculation`; two instances of one class are refused; a device without DHW grows no DHW target; the scenario JSON round trip writes hand-made targets, skips grown ones and refuses declarative names.
- Hygiene: one `_register_output` behind both registration methods, private names for the internal attribute and helper, docstrings that state the rule in one sentence and name `dynamic_components`' generic prefix as the counter-example, the `connect_automatically=False` note, a test matching a port name rather than prose.
- `HISTORY.rst` records the port rename and the vanished phantoms as a result-column change.

Declined: a deprecation shim for the rename (P4 intended-break policy), instance-based naming (the declarative format already does that; the legacy path keeps prefix + weight as F-1 decided), flattening the target-output dataclass, and an attested red-before run (the old constructor demonstrably created the ports).

Checks: 101 targeted tests, three EMS twins unchanged, `household_heatpump_building_sizer / one_week_60s` OK in Python and YAML mode, pylint critical-only 10/10, prospector, flake8 unchanged, mypy (both profiles).

PR #721, opened 2026-09-12
The scenario name and the description travel with the run, not through a global

noah80:
Step 3 of 3 retiring the process-global `SingletonSimRepository`: the last two live keys were run metadata and become run attributes.

`Simulator.scenario_name` and `Simulator.description` replace `RESULT_SCENARIO_NAME` and `DESCRIPTION`. The twelve building-sizer setups assign the scenario hash to the simulator they are handed; `hisim_main` sets the description from the module docstring; the declarative executor sets both from the energy-system file's `name` and `description`. Both travel into `PostProcessingDataTransfer` and the postprocessor reads them there. Fourteen files lose their singleton imports; the two enum members go.

Declarative runs therefore carry a scenario name for the first time: the pyam scenario column reads `household_gas_building_sizer` where it was empty. The Python path's value is unchanged.

Golden-neutral: `household_gas_building_sizer / one_week_60s` passes in Python and YAML mode; twin unchanged. Checks: postprocessing, main, executor, description and building-sizer tests (101 passed, 2 new); flake8 (no new findings), prospector, mypy (both profiles).

After #719 and #720 this leaves the module empty apart from `SingletonMeta`, which `result_path_provider` borrows; the final removal relocates that and drops the module from `hisim/caching/keys.py`'s forbidden-prefix list.


noah80:
Review follow-up, commit cd4c91e5. Taken: literals instead of mirror assertions and a real two-run test (two declarative builds in one process, each transfer object carries its own name and description); the hermetic base test now drives `basic_household` on a predefined profile instead of the LPG-driven sizer; the option-test framework's clone forwards both fields; the postprocessor reads name and description off the transfer object where it uses them, so `scenario.json` no longer depends on a second option being set, and its `name` comes from the same source as the pyam column (filename only as fallback); a new consumer test runs `PostProcessor.run` and asserts the pyam scenario column and `scenario.json`; wording fixes in the executor, simulator and singleton docstrings; the no-op f-strings in the twelve setups.

Two owner decisions: a grouped energy-system file appends its selected variants to the scenario name (`household_gas_building_sizer [electricity_management=ems_with_battery]`), and a Python run that names no scenario is named after its module file. Both in `HISTORY.rst`, with the removed singleton keys.

Declined: deleting the empty enum and class (that is #725, stacked here), a shared helper for the twelve setups, a deprecation shim.

Checks: executor, main, postprocessing, grouping and repository tests (only the known LPG and graphviz environment failures), gas sizer golden in Python and YAML mode, twins and schema unchanged, pylint critical-only, prospector incl. `system_setups/`, flake8 unchanged, mypy (both profiles).

PR #723, opened 2026-09-12
The CHP is a package: component and controller side by side

noah80:
Rule (2026-09-12): one component plus its helpers is one directory under `hisim/components/`, as #578 did for the building.

`generic_chp.py` and `controller_l1_chp.py` become `hisim/components/generic_chp/` with `chp.py`, `controller.py` and an `__init__` re-exporting the six public names. Pure move: the controller is byte-identical, the component changes five lines for the new import path. No recorded twin and no schema entry names either module (neither config declares a preset or constructor), and nothing outside the package imported the controller, so its module path is free to change and no shim is added.

Checks: `test_generic_chp` and `test_economics_extraction` (100 passed), twin freshness on two setups, schema byte-identical, `describe` resolves both classes, pylint critical-only 10/10, prospector, flake8 unchanged, mypy (both profiles), sphinx build.


noah80:
Review follow-up, commit bde62fd2. Taken: `SimpleCHP` and `L1CHPController` pin their module to the package, as the weather and PV packages do, so `get_full_classname()` gives `hisim.components.generic_chp.<Class>` and the `__init__` docstring's stability claim is true (configs keep their submodule spelling, as in the sibling packages; the schema scan lists both); module metadata only in `__init__` (the building convention), merged from both former modules; the totality test in `test_economics_extraction.py` now walks into component packages (twelve classes newly swept, none missing a cost row); README names `generic_chp/chp.py` and states the layout rule (one component and its helpers are one directory); the controller wording (electricity deficit only when a signal is connected); the duplicated prose class list dropped; `HISTORY.rst` records the removed `controller_l1_chp` module path.

Declined: a shim for the old module path (intended break), deriving the test's module map from `__module__`, an auto-derived `__all__`, an `__init__.py` fallback for package-valued map entries, and dropping the controller from the package's public names (the controller is a helper of the component, so it belongs to the package's surface).

Checks: 149 tests across the CHP, economics, schema and record suites; schema unchanged; describe resolves; pylint critical-only 10/10; prospector; flake8 unchanged; mypy (both profiles); sphinx build.

PR #724, opened 2026-09-12
The weather series is the first cache producer, keyed by its own code

noah80:
Phase 1 of the cache service (`roadmap/cache_service_spec.md` §3, §10), first producer.

The Weather's cached computation (six readers, resampling, daily average, DNI) moves into `hisim/components/weather_calculation.py` as a pure producer with a frozen DTO; the component keys it through `CacheKey.for_producer` and the caching client (new `CacheClient.lookup_producer`). Key material: data source, source file content hash, year, timestep, coordinates, and the simulated span only for the DWD 15-minute reader. The producer's import closure is two modules plus numpy/pandas/pvlib, pinned by a test; `ProducerLayering.check` passes.

Why now: on #628 a change to the DNI computation was invisible to the golden gate because the cached frame is keyed on config and simulation parameters only and CI restores the cache from the previous run. A producer's key carries a fingerprint of its own source, so any edit that can change the frame changes the key, in CI and locally. PV series and building solar gains follow as the next two producers; #628 rebases on top.

Bit-identical: `basic_household / one_week_60s` cold (miss logged), warm (hit logged) and YAML mode, plus `air_conditioned_house` (NSRDB), all pass; the twelve series match a copy of the old module value for value across six source/resolution combinations. Twins and schema unchanged. Checks: targeted tests plus `pytest -m base` (env-only failures only), pylint critical-only, prospector, flake8, mypy (both profiles).

First CI run recomputes every cache once (the caching package changed, so the cache action's lineage restarts).


noah80:
Review response, two new commits on top of the three reviewed ones (the branch is also rebased onto main after #721 and #723).

**Acted on**
- The package facade is lazy (PEP 562 `__getattr__`). Importing `hisim.components.weather.calculation` alone now loads nine modules, none of them `hisim.component` or postprocessing; a subprocess test pins that. The six component names still resolve from the package for twins, JSON and `describe`. The four producer-only names left `__all__`; HISTORY records where the seven readers live now.
- `CacheKey.for_producer` stopped rescanning the installed environment: 276 ms to 7 ms per key after the first call in a process.
- The two docstrings give the real reason for the enum's placement: the config's import closure is 45 modules, all hashed into every key.
- The tautological and `dataclasses.replace` tests are replaced by a span test (one day vs one week, DWD_TRY: same digest; DWD_15MIN: differs) and a value pin on temperature, GHI, DHI, wind, azimuth and the daily average at Aachen 2021. DNI and zenith are deliberately not pinned because #628 moves them.
- The DTO enforces its span invariant in `__post_init__` both ways.
- Reader and producer share one frame schema; a missing column raises naming the column and the file instead of zero-filling. `DryBulb` and the chunked reader are gone; the station header is one helper.
- `MODELS_NO_DEVICE = True` replaces the KPI override.
- The daily average is one vectorised block mean (18.2 s to 23 ms on the minutely year, bit-identical to the loop on every span shape tested) and refuses a timestep of two days or more, which used to yield NaN.

**Not changed**
- `get_cost_opex` / `get_cost_capex` overrides stay for now; three PRs stack on this one.
- The `ChainedAssignmentError` warning is the #628 clamp and is fixed there.

Because the producer source changed, the weather key changed again and the first CI run is a cold cache.

PR #727, opened 2026-09-13
The building's solar gains are the third cache producer, chained to the weather's key

noah80:
Third cache producer (spec §3, §10), stacked on the weather producer branch beside the PV one.

The building's solar heat gain through the windows becomes a pure producer in `building/solar_gains.py`, computed up front for the simulated span from the weather's published series. Key material: the weather artifact key, year, timestep, span, and the window geometry as plain values (tilt, azimuth, area, frame fraction, glass transmittance, the two reduction factors); the six weather series are payload. The in-loop computation and the last-timestep cache write are gone, which also closes the deferred-write defect the spec's survey lists. The lookup happens at the first simulate call rather than at prepare, because 16 of 18 setups add the building before the weather.

Finding worth knowing: the legacy solar-gains cache was never read. The guard tested one attribute name and the cached series was assigned to another, so every run recomputed in the loop and the written file was dead. A value-for-value comparison of the produced series against a legacy cache file gives a maximum difference of 0.0 over 10 080 values.

Goldens: `basic_household / one_week_60s` cold (miss logged), warm (hit), YAML mode, plus five further week pairs including a NSRDB weather, all OK. Twins and schema unchanged. Tests: the new producer suite (chain, fingerprint, closure), the building suites, weather and caching suites; pylint critical-only, prospector, flake8 unchanged, mypy (both profiles).


noah80:
Closed by GitHub when the base branch was deleted; continued in the PR above against main, same branch.

PR #728, opened 2026-09-13
The electricity meter's legacy factory goes; every caller uses the preset

noah80:
Deletes `ElectricityMeterConfig.get_electricity_meter_default_config` (17 setups, 10 test files); every caller uses `preset_standard`. The factory's only override user passes the name to the preset instead.

### DIFF
--- a/energy_systems/air_conditioned_house.energy_system.yaml
+++ b/energy_systems/air_conditioned_house.energy_system.yaml
@@ -152,12 +152,7 @@ components:
       - Building
   ElectricityMeter:
     class: hisim.components.electricity_meter.ElectricityMeter
-    config:
-      device_co2_footprint_in_kg: null
-      investment_costs_in_euro: null
-      lifetime_in_years: null
-      maintenance_costs_in_euro_per_year: null
-      subsidy_as_percentage_of_investment_costs: null
+    preset: standard
     inputs:
       - from: AirConditioner.ElectricalPowerConsumption
         tags:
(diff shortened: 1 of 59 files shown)


noah80:
Review response, one new commit here and one at the top of the stack (#731).

**Acted on**
- The building-override test asserted what `dataclasses.replace` did, not what the preset did; deleted. The defaults test was subsumed by the name-passthrough test; deleted. Six tests remain in the file.
- `preset_naming_supplement.md` no longer lists the deleted factory; the spelling count is corrected (four `get_<x>_default_config` factories remain in the tree).
- The narrative lines of `p4_component_sweep_requirements.md` (lines 18, 22, 27) also named the boiler-controller, HDS-controller and UTSP factories that #729 to #731 delete, so they are rewritten once, on #731, to be true after the whole stack merges. The class survey is a dated 2026-08-27 snapshot and stays as it is.

**Intended, not changed**
- No deprecation shim: the sweep's policy is a clean break, and every in-repo caller moved.
- No `component_id` or building parameter on `preset_standard`: the `@preset` contract fixes the signature to the name. Nothing live sets a building on a meter (the declarative loader injects the name only, no setup passes one, the district setups live in `obsolete/`), and the district branch in the component is pre-existing code this PR does not touch.
- `builder_args` in the economics contract test: PV and battery still have zero-argument legacy factories while the meter has a name-taking preset; the tuple goes when those two convert.
- The literal `"ElectricityMeter"` at the call sites is the fleet rule: the instance name is the caller's, and the declarative path names a component by its YAML key.

PR #729, opened 2026-09-13
The boiler controller's four legacy factories go; presets and explicit overrides replace them

noah80:
Deletes the four `GenericBoilerControllerConfig` factories (8 setups, 3 tests). Modulating callers use `preset_modulating` resolved from the boiler's power; on/off, pellet and wood-chip use `preset_on_off`, the last two with the exact runtimes the factories set (30/15 and 60/30 minutes) as explicit overrides. Component names stay as the preset's name argument, so no recorded key moved.


noah80:
Review response, one new commit.

**Acted on**
- The two presets are pinned to the four factories they replaced: a base test resolves `preset_modulating` and `preset_on_off` against a sizing context and asserts every field the factories wrote (modulating 1800/1800, on/off 0/0, 5.0 K full-power difference, 10 K hysteresis, 16 °C threshold, no secondary mode, no DHW). A second test pins the pellet and wood-chip timings (30/15 and 60/30 minutes) through the recorded twins, which the freshness gate ties to the setups.
- The mode test's parametrized threshold and hot-water flag now reach the controller; the helper had hard-set one and ignored the other. No expectation changed.

**Intended, not changed**
- `preset_on_off` reproduces the deleted generic on/off factory exactly, zero runtime and zero resting time included; `preset_modulating` resolves to the modulating factory's literals. The presets are byte-faithful to what they replace.
- Pellet and wood chip as `on_off` plus two field overrides is the recorded P4 plan for this class: the fuel is not a second control law, it moves two numbers. The values are now pinned by test.
- `with_domestic_hot_water_preparation` stays a field override with default `False`: boolean flags are overrides rather than preset axes, and whether a system prepares hot water is a property of the system, not of the control law.
- No deprecation shim, by the sweep's policy; every in-repo caller moved.
- The naming supplement's row for this class is rewritten at the top of the stack (#731) together with the other B1 rows; the class survey is a dated snapshot and stays.

PR #731, opened 2026-09-13
The occupancy's legacy factory goes; every caller uses the preset

noah80:
Deletes `UtspLpgConnectorConfig.get_default_utsp_connector_config` (18 setups, 17 test files); all 16 preset fields equal the factory's. The 11 building sizers keep their acquisition mode, household and cache directory as field assignments on the preset rather than the `for_household` constructor, because the constructor would also null the predefined-profile name, which is recorded in the twins and the car sizer's occupancy identity: a value change that belongs in its own commit, noted at each site and in R3.


noah80:
Closed by GitHub when the base branch was deleted; continued in the PR above against main, same branch.

